PR #22, opened 2021-07-09
add a github actions task to check wasm builds work

willemolding:
I decided to see if our current codebase supports building to Wasm and it does! It is important we keep it this way and avoid adding dependencies or patterns that Wasm doesn't support. Future PRs will fail to pass CI if they do not support building to Wasm.

**Summary of changes**
- Adds a github action task that installs the Wasm target and runs cargo check against it

Based off suggestion here: https://rustwasm.github.io/docs/book/reference/add-wasm-support-to-crate.html#maintaining-ongoing-support-for-webassembly

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -62,3 +62,19 @@ jobs:
         with:
           command: clippy
           args: -- -D warnings
+
+  wasm:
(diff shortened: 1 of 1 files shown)


creativcoder (approved):
lgtm!

PR #25, opened 2021-07-14
Willem/layout parser

willemolding:
**Summary of changes**
- Adds a library for parsing type layout files as defined here https://github.com/MinaProtocol/mina/issues/8718
- Adds CLI for reading these from json files and printing

This essentially allows us to have runtime access to the type trees for OCaml types. These can be generated for any type in the OCaml Mina implementation that supports serialization.

What we do with this is still to be decided. It could be used for code generation, manipulation of serialized binary files or test generation.


layout_parser/src/bin/cli.rs
```diff
@@ -0,0 +1,42 @@
...
+    let file = File::open(path)?;
+    let reader = BufReader::new(file);
+
+    // Read the JSON contents of the file as an instance of `Layout`.
+    // need to use the disabled recursion limit for large layouts (e.g. external_transition)
+
+    let mut deserializer = serde_json::Deserializer::from_reader(reader);
+    deserializer.disable_recursion_limit();
+    let deserializer = serde_stacker::Deserializer::new(&mut deserializer);
+    let u = Layout::deserialize(deserializer)?;
```
noot, line 32: it looks like this just deserializes the layout from JSON, but it doesn't parse it into the bin_prot types unless I'm missing something? would be good to get that parsing in, but happy to approve this and that can be done in a follow up

creativcoder: Follow https://github.com/MinaProtocol/mina/issues/8718 for developments on this. It would be great if we merge this after implementing the ability to generate Rust types or conclude if it isn't possible after improvements in the layout JSON.

willemolding: @creativcoder sure sounds good. This PR just includes the ability to deserialize the layouts into a type tree but we really need to decide what we want to do with that.

PR #28, opened 2021-07-15
implement common consensus algorithms

### LINKED ISSUE #23: implement common consensus algorithms
**Task summary**

- implement the common consensus algorithms outlined here: https://github.com/jspada/mina/tree/8e6558e5f698cb01f5f99a45dc097d04729f7128/docs/specs/consensus#3-algorithms
- will require the `protocol_state` type to be implemented (analagous to block header), in particular the `consensus_state` field. the other fields don't appear to be necessary for the common consensus algorithms

noot:
**Summary of changes**
- implement common consensus algorithms for `ProtocolStateChain` which is of type `Vec<ProtocolState>`

would love feedback on this in terms of  how to make my code more idiomatic and if the types I created for consensus make sense, in particular the traits and usage of generics, and if there's a better way to create a `Chain` type that's an ascending list of `ProtocolState`s. 

looking forward we will likely need to have a blocktree of all the current forks in the network, so maybe it would be nice to have a blocktree implementation and create chains from that.

Closes #23


types/src/protocol_state.rs
```diff
@@ -0,0 +1,107 @@
...
+use mina_crypto::hash::BaseHash;
+use serde::{Deserialize, Serialize};
+
+// TODO: confirm below types
+pub type Length = u64;
+pub type VrfOutput = [u8; 32];
+pub type Amount = u64;
+
+/// GlobalSlot contains (slot_number, slots_per_epoch)
+pub type GlobalSlot = (u64, u64);
```
creativcoder, line 13: Quite a few of these types are already defined in the base crate: https://github.com/ChainSafe/mina-rs/blob/main/base/src/global_slot.rs We can just import and use them.

consensus/src/common.rs
```diff
@@ -0,0 +1,88 @@
...
+            None => return None,
+        };
+
+        let mut output = Vec::<u8>::new();
+        to_writer(&mut output, &s).unwrap();
+
+        let mut hasher = Blake2b::new();
+        hasher.update(output);
+
+        // TODO: is there a prettier way to do this?
```
creativcoder, line 82: It would be good to add an impl for BaseHash from a slice:
```rust
// In crypto/hash/types.rs

impl<'a> From<&'a [u8]> for BaseHash {
    fn from(b: &'a [u8]) -> Self {
        let mut o = BaseHash::default();
        o.0.copy_from_slice(&b);
        o
    }
}

```
and then we can do:
```
let hash = hasher.finalize();
let hash = hash.as_slice(); // This temporary is just to make rust borrow checker happy
Some(BaseHash::from(hash))
````

File: types/src/protocol_state.rs (hunk 1 above)
Comment (noot) [reply types/src/protocol_state.rs:13]: wow I totally missed this, will update

File: base/src/protocol_state.rs [hunk 3]
```diff
@@ -39,20 +39,30 @@ pub struct ProtocolConstants {
...
-    consensus_state: ConsensusState,
+    pub consensus_state: ConsensusState,
     constants: ProtocolConstants,
 }
+
+pub trait Header {
+    fn get_height(&self) -> Length;
+}
+
+impl Header for ProtocolState {
```
creativcoder, line 64: Is `ProtocolState` the only implememter for `Header`?
What are other entities that can implement `Header`?

consensus/src/common.rs
```diff
@@ -0,0 +1,211 @@
...
+use serde_bin_prot::to_writer;
+use std::convert::TryInto;
+
+pub struct ProtocolStateChain(Vec<ProtocolState>);
+
+pub trait Chain<T>
+where
+    T: Header,
+{
+    fn push(&mut self, new: T) -> Result<(), &'static str>;
```
creativcoder, line 18: The second type param of Result needs to be a proper error type. We can define and enum ConsensusError which derives thiserror::Error from https://docs.rs/thiserror/1.0.26/thiserror/ crate use it in place of &'static str

consensus/src/common.rs
```diff
@@ -0,0 +1,211 @@
...
+    fn state_hash(&self) -> Option<BaseHash>;
+}
+
+impl Chain<ProtocolState> for ProtocolStateChain {
+    fn push(&mut self, new: ProtocolState) -> Result<(), &'static str> {
+        match self.0.len() {
+            0 => (),
+            n => {
+                if new.get_height().0 != self.0[n - 1].get_height().0 + 1 {
+                    return Err("header must have height 1 greater than top");
```
creativcoder, line 34: can probably use an enum variant ConsensusError::InvalidHeight along with changes from the above comment.

consensus/src/common.rs
```diff
@@ -0,0 +1,211 @@
...
+        self.0.push(new);
+        Ok(())
+    }
+
+    fn top(&self) -> Option<&ProtocolState> {
+        self.0.last()
+    }
+
+    fn consensus_state(&self) -> Option<&ConsensusState> {
+        match self.top() {
```
creativcoder, line 48: We can replace the whole match with just `self.top().map(|s| &s.body.consensus_state)`

consensus/src/common.rs
```diff
@@ -0,0 +1,211 @@
...
+            Some(s) => Some(&s.body.consensus_state),
+            None => None,
+        }
+    }
+
+    fn global_slot(&self) -> Option<&GlobalSlot> {
+        match self.top() {
```
creativcoder, line 55: ditto as above

base/src/protocol_state.rs
willemolding: Even if no other types do implement it I like this for separation of concerns

consensus/src/common.rs
willemolding: Seconded, check out https://github.com/ChainSafe/serde-bin-prot/blob/main/src/error.rs for an example of writing a custom error type

consensus/src/common.rs
```diff
@@ -0,0 +1,211 @@
...
+
+        let hash = blake2b(
+            32,
+            &[],
+            &s.body.consensus_state.last_vrf_output.0.as_bytes(),
+        );
+        BaseHash::from(hash.as_bytes()).to_hex()
+    }
+
+    fn state_hash(&self) -> Option<BaseHash> {
```
willemolding, line 84: It is important to use the correct hash types. By convention Mina adds a prefix string to data before hashing it which is different depending on the type being hashes. Presumably this is to prevent a hash collision between different types with identical data. 

Check out the prefixes here https://github.com/ChainSafe/mina-rs/blob/8daf8e2a45e19c3b4912c688744faaa20cee6388/crypto/src/hash/prefixes.rs#L76

To avoid confusion I made the `Hash` and `Hashable` traits. Here `ProtocolState` should implement `Hashable` which returns a `StateHash` which implements `Hash`. This also makes these parts of the code much tidier since all the hashing logic will live in once place.

This is a bit tricky to describe in a comment so I'll link a branch showing how this could be done using our hashing lib

crypto/src/hash/types.rs
```diff
@@ -13,14 +13,15 @@
...
+use std::fmt::Write;
 
 use serde::{Deserialize, Serialize};
 use serde_versions_derive::version;
 
 pub(crate) type HashBytes = Box<[u8]>;
 
 #[derive(Default, Serialize, Deserialize, PartialEq, Debug, Clone)]
-pub(crate) struct BaseHash([u8; 32]);
+pub struct BaseHash([u8; 32]);
```
willemolding, line 24: I'd be happier to keep this as in internal crate type if possible

crypto/src/hash/types.rs
```diff
@@ -31,10 +32,28 @@ impl From<HashBytes> for BaseHash {
...
+impl BaseHash {
+    pub fn to_hex(&self) -> String {
+        let mut s = String::from("0x");
+        for b in &self.0 {
+            write!(&mut s, "{:02x}", b).expect("failed to write hex byte")
+        }
+        s
+    }
+}
+
```
willemolding, line 52:
```suggestion
impl AsRef<[u8]> for BaseHash {
    fn as_ref(&self) -> &[u8] {
        &self.0
    }
}
```

and then if we use the `hex` crate this type will automatically implement `to_hex` whenever the trait is in scope!

willemolding (changes requested):
Amazing job overall. Stoked to be making progress on the next milestone.

A few changes suggested around the use of the hashing lib. I've left some comments in here but to make it easier I have a branch with some of the changes suggested which I think will be much easier to follow.

It also required some changes to the hash lib so it was great to have that tested out. 

Some suggested changes are in here https://github.com/ChainSafe/mina-rs/tree/noot/common-willem-rev

consensus/src/common.rs
```diff
@@ -0,0 +1,211 @@
...
+            None => None,
+        }
+    }
+
+    fn epoch_slot(&self) -> Option<u32> {
+        self.global_slot()
+            .map(|s| (s.slot_number.0 % s.slots_per_epoch.0))
+    }
+
+    fn length(&self) -> u64 {
```
creativcoder, line 66: Better to use a `usize` type as it's a generic type which changes based on architecture. On a 64 bit it'll be a u64 and likewise for 32bit.

consensus/src/common.rs
```diff
@@ -0,0 +1,211 @@
...
+        let hash = blake2b(
+            32,
+            &[],
+            &s.body.consensus_state.last_vrf_output.0.as_bytes(),
+        );
+        BaseHash::from(hash.as_bytes()).to_hex()
+    }
+
+    fn state_hash(&self) -> Option<BaseHash> {
+        let s = match self.top() {
```
creativcoder, line 85: It would be great to use if let Some(s) = self.top() here as match is best applicable for more than 3 match arms.

consensus/src/common.rs
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    fn length(&self) -> u64 {
+        self.0.len().try_into().unwrap()
+    }
+
+    fn last_vrf(&self) -> String {
+        let s = match self.top() {
+            Some(s) => s,
+            None => return "0x".to_string(),
```
creativcoder, line 73: Is this correct? I guess we should return a `None`, if the the chain is empty.

creativcoder (changes requested):
Thank you for giving this a head start! Left a few comments

base/src/protocol_state.rs
noot: @creativcoder right now it's just `ProtocolState` but potentially `ExternalTransition` could implement it as well

consensus/src/common.rs
noot: done

consensus/src/common.rs
noot: done, I used an else case to return `None`:
```
        if let Some(s) = self.top() {
            let mut output = Vec::<u8>::new();
            to_writer(&mut output, &s).unwrap();
            let hash = blake2b(32, &[], &output);
            Some(BaseHash::from(hash.as_bytes()))
        } else {
            None
        }
```

is this the right way to do this?

consensus/src/common.rs
noot: updated to return `Option<String>`

consensus/src/common.rs
noot: updated

consensus/src/common.rs
noot: updated

consensus/src/common.rs
noot: updated

consensus/src/common.rs
noot: updated!

crypto/src/hash/types.rs
noot: oh nice, so by implementing `AsRef` for a type, then  `to_hex` automatically gets implemented?

noot:
@willemolding thanks for the updates on your branch, I pulled them into here, much cleaner to use the `Hashable` trait :) 

@creativcoder addressed your comments!

crypto/src/hash/types.rs
willemolding: Yep you got it As long as the trait is in scope that is

consensus/src/common.rs
willemolding: You could use `map` here as well since you are only changing what is contained in the `Some` variant
```rust
self.top().map(|s| {
    let mut output = Vec::<u8>::new();
    to_writer(&mut output, &s).unwrap();
    let hash = blake2b(32, &[], &output);
    BaseHash::from(hash.as_bytes())
})
```

crypto/src/hash/traits.rs
```diff
@@ -13,7 +13,7 @@ const BLAKE_HASH_SIZE: usize = 32;
 /// This defines the prefix that is added to the data prior to it being hashed
 pub trait Hash
 where
-    Self: From<Box<[u8]>>,
+    Self: From<Box<[u8]>> + AsRef<[u8]>,
```
willemolding, line 16: Does this need an implementation of `AsRef` as a trait bound? It would be nice if we could drop this from here so that types can still implement `Hash` either way and just need to implement `AsRef<[u8]>` if they need converting to hex.

willemolding (approved):
Looks amazing! One tiny comment but that's all

consensus/src/common.rs
willemolding: DW looks like you are all over this

crypto/src/hash/traits.rs
```diff
@@ -60,6 +60,12 @@ mod tests {
         const PREFIX: &'static HashPrefix = PROTOCOL_STATE;
     }
 
+    impl AsRef<[u8]> for TestHash {
```
creativcoder, line 63: I guess TestHash was for testing and can be removed?

crypto/src/hash/types.rs
```diff
@@ -31,10 +30,24 @@ impl From<HashBytes> for BaseHash {
     }
 }
 
+impl<'a> From<&'a [u8]> for BaseHash {
```
creativcoder, line 33: Do we still need this impl, after implementing `AsRef` below?

creativcoder (approved):
Neat! let's merge this soon post last set of changes.

crypto/src/hash/types.rs
noot: I think the other hash implementations still need it, for example this doesn't work if there's no `AsRef` for `BaseHash`:
```
pub struct StateHash(BaseHash);

impl AsRef<[u8]> for StateHash {
    fn as_ref(&self) -> &[u8] {
        &self.0.as_ref()
    }
}
```

crypto/src/hash/traits.rs
noot: it's used in the test cases here, so would be good to keep it

crypto/src/hash/traits.rs
noot: removed

PR #37, opened 2021-08-20
ADD initial account database work

mmacedoeu:
**Summary of changes**
- 
Milestone 4
Store and Update account database
Merkle-ledger

Closes 

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,6 +21,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
+version = "1.0.43"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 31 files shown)


willemolding:
Hey Marcos, Just had a quick browse through this. Keep in mind this is a reimplementation of the protocol and not a direct port of the OCaml code. So don't feel like you need to be constrained to use the same module structure or type names.

For example I don't think we should follow the `<module-name>/stable/v1` or `intf/<interface-name>` naming conventions. These might be idiomatic OCaml but I don't think they are so much in Rust (especially since there are no interfaces :P).

mmacedoeu:
The intention here is to kick start and split work into it's own issues, it is heavily inspired by original ocaml code for now just to start things, should be completely refactored, M4 at this stage don't have a detailed spec

PR #38, opened 2021-08-25
Ability to deserialize block to bin_prot::Value

### LINKED ISSUE #16: Deserialize a block from file to our types

### LINKED ISSUE #33: Implement serialization for types that use custom strategy
**Task summary**

Some types used in Mina inside of `external_transition` do not use the bin_io method for serialization. They instead use a custom serialization strategy.

These types are:
- "Ledger_hash0",
- "State_hash",
- "Pending_coinbase.Stack_hash",
- "State_body_hash",
- "Pending_coinbase.Hash_builder",
- "Snark_params.Make_inner_curve_scalar",
- "Snark_params.Tick",
- "Epoch_seed",
- "Zexe_backend.Zexe_backend_common.Stable.Field",
- "Vector.Vector_18",
- "Vector.Vector_4",
- "Pending_coinbase.Coinbase_stack",
- "Vector.Vector_2"

For each of these we need to figure out how they can be serialized/deserialized in order to support them.

There is also a need to add support to serde-bin-prot when deserializing in loosely typed mode using a layout file to support custom types. This can be done by passing a map of names (see above) to functions with the signature 
```rust
fn<'de, D>(D) -> Result<bin_prot::Value, D::Error> where D: Deserializer<'de>
```

willemolding:
**Summary of changes**
- Fixes a bug in how option types were handled in `deserialize_loose.rs`
- Adds new tests for option types
- Updates the block layout with the fixed one provided by O(1)
- Properly handles custom BigInt types
- Tidies up loose deserializer code
    - Removes use of `return`
    - Adds strongly typed errors
    - Removes use of println! for debugging
- Adds a test showing deserialization of a block and access to its data!

Closes #33
Closes #16 


bin-prot/tests/layouts.rs
```diff
@@ -162,6 +189,13 @@ fn smoke_test_deserialize_block() {
...
-    // This is failing at the moment so don't check the result
+    let block: Value = Deserialize::deserialize(&mut de).expect("Failed to deserialize block");
+
+    assert_eq!(
+        block["t"]["protocol_state"]["t"]["t"]["previous_state_hash"]["t"],
+        Value::String(vec![
+            30, 76, 197, 215, 115, 43, 42, 245, 198, 30, 253, 134, 49, 117, 82, 71, 182, 181, 180,
+            95, 18, 250, 46, 1, 25, 3, 78, 193, 57, 152, 116, 49,
+        ])
+    );
```
noot, line 200: do we want to check for the other fields in the block as well in this test? also, is there a way to get around the `["t"]`s so we can directly access `block["protocol_state"]["previous_state_hash"]`, or not really due to how the deserializer works?

noot (approved):
amazing work! just one comment, but it's not blocking

bin-prot/tests/layouts.rs
willemolding: I was thinking I'll write something that flattens the values by squashing out the `{version, t}` parts. In a future PR!

PR #52, opened 2021-09-28
Willem/wire types derive macro

willemolding:
**Summary of changes**

- Adds two new crates wire-type and wire-type derive
- These replace serde_versions_derive as the preferred way to add version numbers to structs
- Updates types using serde_versions_derive to use wire-type instead

The improved implementation allows the derive macro
`#[derive(WireType)]` to be added to structs or enums. The macro automatically generates a corresonding wire type
which includes a version number (by default this is 1 but can be set using `#[wire_type( version = 2 )]`. 

It is also possible to have types automatically use their own version WireType whenever they are serialized or deseralized using serde. This makes it much easier to manage which types are using versions and which are not. This is done be using the serde `from` and `into` attributes
```rust
  #[serde(from = "<Self as WireType>::WireType")] // can be deserialized from its wire type
   #[serde(into = "<Self as WireType>::WireType")] // will be serialized to its wire type
```


consensus/src/common.rs
```diff
@@ -167,17 +167,17 @@ mod tests {
         assert_eq!(epoch_slot, Some(2));
     }
 
-    #[test]
-    fn test_protocol_state_chain_state_hash() {
-        let mut c: ProtocolStateChain = ProtocolStateChain(vec![]);
+    // #[test]
```
creativcoder, line 170: Are these commented accidentally?

creativcoder (approved):
LGTM, nice upgrade to versions derive!

consensus/src/common.rs
willemolding: Oh yeah they are, There should be no changes to this file

wire-type-derive/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+            fn from_wire_type(t: Self::WireType) -> Self { Self::from(t) }
+        }
+
+    };
+    output.into()
+}
+
+// Creates the token stream for a shadow struct/enum for the input
+// The shadow struct has an identical structure but its own serialize/deserialize
+// implementations. This is to prevent the infite recursion if using serde from and into
```
noot, line 74:
```suggestion
// implementations. This is to prevent the infinite recursion if using serde from and into
```

wire-type/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+            b: u32,
+        }
+
+        let x = X { a: 123, b: 321 };
+
+        let serialized = serde_json::to_string(&x).unwrap();
+        println!("serialized = {}", serialized);
+
+        let deserialized: X = serde_json::from_str(&serialized).unwrap();
+        println!("deserialized = {:?}", deserialized);
```
noot, line 144: should this assert that `x` and `deserialized` are the same?

wire-type/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+        #[serde(into = "<Self as WireType>::WireType")] // will be serialized to its wire type
+        struct X(u32, u32);
+
+        let x = X(123, 321);
+
+        let serialized = serde_json::to_string(&x).unwrap();
+        println!("serialized = {}", serialized);
+
+        let deserialized: X = serde_json::from_str(&serialized).unwrap();
+        println!("deserialized = {:?}", deserialized);
```
noot, line 160: same here

noot (approved):
great work :D

PR #53, opened 2021-09-30
Add serialization support for bin_prot::Value

willemolding:
In order to do round-trip for the subtypes of block we need the following:

- The ability to serialize a loose type bin_prot::Value back to its binary from
- The ability to deserailize into a type that contains a mix of bin_prot::Value fields and strongly typed fields

This PR adds both of these

**Summary of changes**
- Modified bin_prot deserialize to work on hybrid types. (e.g. `struct S { a: i32, b: bin_prot::Value }`) This requires ensuring that if a type layout is provided then all deserialize calls get routed to `deserialize_all` so the layout can be correctly traversed
- Adds ability to serialize a `bin_prot::Value` type back to binary form.
- Adds tests showing round-trips for loose types


bin-prot/src/value/layout/traverse.rs
```diff
@@ -62,23 +47,19 @@ pub enum Error {
...
         match top {
             Some(rule) => {
                 match rule {
-                    BinProtRule::List(r) => {
-                        // the code driving the iterator should call `repeat` if it encounters a list
-                        self.stack.push(*r);
+                    BinProtRule::List(_rule) => {
+                        // the code driving the iterator should take the list rule, push it
+                        // then call repeat the required number of times
+                        // return Some(Self::Item::List(*rule))
```
willemolding, line 62:
```suggestion
```

bin-prot/src/value/layout/traverse.rs
```diff
@@ -87,28 +68,31 @@ impl BranchingIterator for BinProtRuleIterator {
...
+                        unimplemented!();
+                        // polyvars
+                        //     .into_iter()
+                        //     .map(|s| match s {
+                        //         Polyvar::Tagged(pv) => pv.polyvar_args,
+                        //         Polyvar::Inherited(rule) => {
+                        //             vec![rule]
+                        //         }
+                        //     })
+                        //     .collect()
```
willemolding, line 95:
```suggestion
```

bin-prot/src/value/mod.rs
```diff
@@ -26,11 +26,11 @@ pub enum Value {
     Nat0(u32),
     Bool(bool),
     String(Vec<u8>),
-    Char(char),
+    Char(u8),
     Int(i64),
     Float(f64),
     Option(Option<Box<Value>>),
-    Record(HashMap<String, Value>), // records/structs
+    Record(Vec<(String, Value)>), // records/structs
```
willemolding, line 33: Important lesson here. Never use `HashMap` for something that needs deterministic ordering of fields. This led to flaky failing tests because the ordering would randomly change

bin-prot/tests/layouts.rs
```diff
@@ -165,37 +191,125 @@ fn test_option_rule() {
...
+                180, 95, 18, 250, 46, 1, 25, 3, 78, 193, 57, 152, 116, 49,
+            ]
+            .iter()
+            .map(|c| Value::Char(*c))
+            .collect()
+        )
     );
+
+    // check roundtrip
+    test_roundtrip(&block["t"]["protocol_state"], &BLOCK_BYTES[1..845]);
```
willemolding, line 285:
```suggestion
    test_roundtrip(&block, &BLOCK_BYTES);
```

bin-prot/src/value/mod.rs
creativcoder: or can use https://docs.rs/indexmap/1.7.0/indexmap/ too

bin-prot/src/value/layout/traverse.rs
```diff
@@ -62,23 +47,18 @@ pub enum Error {
     InvalidBranchIndex { max: usize, got: usize },
 }
 
-impl BranchingIterator for BinProtRuleIterator {
+impl Iterator for BinProtRuleIterator {
```
creativcoder, line 50: Nice to see that we can use the builtin `Iterator` now. Also do we need an update to the module level docs above since this impl has changed slightly?

bin-prot/tests/layouts.rs
```diff
@@ -165,37 +191,125 @@ fn test_option_rule() {
...
+  ]
+]
+"#;
+
+#[test]
+fn test_multiple_ctor_arg_sum_rule() {
+    let rule: BinProtRule = serde_json::from_str(MULTIPLE_CTOR_ARG_SUM_RULE).unwrap();
+    let example = vec![0x00, 0x05, 0x06]; // One({ first: 5 }, { second: 6})
+
+    let mut de = Deserializer::from_reader_with_layout(Cursor::new(example.as_slice()), rule);
```
creativcoder, line 242: nit: can remove the Cursor wrapper as https://doc.rust-lang.org/std/io/trait.Read.html#impl-Read-2

willemolding: While it can be safely removed I think I might leave it but only in the test code. It was useful for debugging since `Cursor` implements `Seek` so you can figure out exactly which input byte it is failing at. It isn't used currently but if we add it back it in would be useful to have in the tests.

PR #62, opened 2021-10-20
Willem/tidy deserializer api

willemolding:
**Summary of changes**
- Uses type state pattern to:
    - simplify the implementation of the bin-prot deserializer
    - Improve its API
    - Move some runtime checks to compile time (woop!) making the de-serialization more efficient

### DIFF
--- a/bin-prot/src/de.rs
+++ b/bin-prot/src/de.rs
@@ -9,9 +9,10 @@ use crate::ReadBinProtExt;
 use byteorder::{LittleEndian, ReadBytesExt};
 use serde::de::{self, value::U8Deserializer, EnumAccess, IntoDeserializer, Visitor};
 use serde::Deserialize;
+use duplicate::duplicate;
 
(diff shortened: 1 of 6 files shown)


bin-prot/src/de.rs
```diff
@@ -7,26 +7,37 @@ use crate::error::{Error, Result};
...
+            mode: StronglyTyped,
         }
     }
+}
 
-    pub fn from_reader_with_layout(rdr: R, layout: &BinProtRule) -> Self {
-        Self {
-            rdr: BufReader::new(rdr),
-            layout_iter: Some(layout.clone().into_iter()),
+impl<R: Read> Deserializer<R, StronglyTyped> {
```
creativcoder, line 34: Shouldn't the 2nd param be `LooselyTyped`?

creativcoder: Got it

Issue #77, opened 2021-11-08
Moving bin prot wire-protocol related crates under /protocol


creativcoder:
**Issue summary**

Proposing a crate re-org effort before closing M1. Current the bin-prot test related code are spread at the root directory. Thinking of having a protocol/ directory at root which would contain everything that has to do with wire protocol and bin prot related.

So here's what the updated workspace Cargo.toml would look like after this refactor:

```
[workspace]
members = [
	"base",
	"consensus",
	"crypto",
    # All crates related to bin prot wire protocol
	"protocol/bin-prot",
	"protcol/wire-type",
	"protocol/bin_prot_checker",
	"protocol/wire-type-derive",
	"protocol/test-fixtures",
	"protocol/test-serialization",
]
```

Accordingly the dependent crates would update their `Cargo.toml` for crates in protocol/

Open to re-org for other crates as well.

Thoughts?

**Other information and links**

polarbeargo:
cargo build & cargo test successful
![Screenshot (234)](https://user-images.githubusercontent.com/8589224/140701922-0fcf0891-bb53-4553-9d8a-e6d1101cc216.png)
![Screenshot (235)](https://user-images.githubusercontent.com/8589224/140701946-8a6cf781-7bc4-4fc8-9515-d970d54039ac.png)
.

PR #78, opened 2021-11-08
Moving bin prot wire-protocol related crates under /protocol #77

polarbeargo:
**Summary of changes**
- 
Now the updated workspace Cargo.toml after this refactor look like the following 
```
[workspace]
members = [
	"base",
	"consensus",
	"crypto",
	# All crates related to bin prot wire protocol
	"protocol/bin-prot",
	"protocol/wire-type",
	"protocol/bin_prot_checker",
	"protocol/wire-type-derive",
	"protocol/test-fixtures",
	"protocol/test-serialization",
]

```

 Include the issue reference this pull request is connected to  #77 
Closes 

Screenshot (234)
Screenshot (235)


Cargo.toml
```diff
@@ -1,12 +1,13 @@
...
-	"bin_prot_checker",
 	"consensus",
 	"crypto",
-	"bin-prot",
-	"wire-type",
-	"wire-type-derive",
-	"test-fixtures",
-	"test-serialization",
+	# All crates related to bin prot wire protocol
+	"protocol/bin-prot",
```
hanabi1224, line 7: I think wildcard should be sufficient here

hanabi1224 (approved):

Cargo.toml
creativcoder: we might also put non-crates such as docs

creativcoder:
We can also put layouts directory within protocol as its scope is only to be used by bin-prot related crates.

Cargo.toml
polarbeargo: How about using [SwaggerHub](https://support.smartbear.com/swaggerhub/docs/enterprise/release-notes.html)?

polarbeargo:
After putting the layouts directory within the protocol folder, cargo build & cargo test successful.
Screenshot (236)
Screenshot (237)

creativcoder (approved):
That was fast! :confetti_ball:

polarbeargo:
@creativcoder thank you for the encouragement! Very appreciate 😊!

willemolding (approved):
I think also at some stage we should move a bunch of the types implementing WireType to the protocol subdir as well. Many of these won't really be used by the core logic and just serve to encode how the deserialization protocol should work.

This looks great though for a start :+1:

Issue #24, opened 2021-07-13
M2 - implement consensus checkpointing algorithms


noot:
**Task summary**

During chain/block selection, checkpointing is an important step in mina consensus that helps in figuring out if a fork is a long range or a short range.

- implement `initCheckpoints`, `updateCheckpoints`, and `isShortRange` algorithms outlined here: https://github.com/MinaProtocol/mina/blob/32f529d8d8d712a44ee75be66061ce08cbdc8924/docs/specs/consensus/README.md#53-decentralized-checkpointing
- requires `consensus_state` type to be defined

polarbeargo:
To avoid  x86_64 & wasm CI error, the function name should be `init_checkpoints`, `update_checkpoints` and `is_short_range`. Just a reminder to myself.

polarbeargo:
**Summary of changes**
Changes introduced in this pull request:
- 
Base on the branch [noot/checkpoint](https://github.com/ChainSafe/mina-rs/commits/noot/checkpoint) continue with issue #24 

- [x] Add `initCheckpoints`, and `isShortRange` algorithms outlined here: https://github.com/MinaProtocol/mina/blob/32f529d8d8d712a44ee75be66061ce08cbdc8924/docs/specs/consensus/README.md#53-decentralized-checkpointing
- [x] Add defined `consensus_state` type.
- [x] Add license in `checkpoint.rs`.
- [x] Update library.
- [x] Fix: Use of moved value.
- [x] Test: Implement Implement wasm-bindgen-test in consensus directly for issue #72 .
- [x] Feat: Implement `is_short_range()` for short-range fork check, `test_init_checkpoints` and , `init_checkpoints` for  genesis checkpoints follow new spec.  

Cargo build & wasm-pack test --node passed.
**Specification reference**

- 
https://github.com/MinaProtocol/mina/blob/32f529d8d8d712a44ee75be66061ce08cbdc8924/docs/specs/consensus/README.md

**Other information and links**

https://github.com/MinaProtocol/mina/blob/develop/src/lib/consensus/proof_of_stake.ml

PR #34, opened 2021-08-04
[M2] Implement chain density and select chain API

creativcoder:
**Summary of changes**

The density APIs are used in case when the adversary creates a long range fork in the network. The idea is to track the minimum density (which is valid during the first few blocks at the point of fork of the adversary chain).

Changes introduced in this pull request:
- Implements APIs described in the spec doc.
- ~Some of the details are missing as per sec 7.4 (page 25) https://eprint.iacr.org/2020/352.pdf~

**Reference issue to close (if applicable)**

Closes #26 

**Other information and links**

PR #80, opened 2021-11-10
M2 - implement consensus checkpointing algorithms 

polarbeargo:
**Summary of changes**
- 
Base on the branch [noot/checkpoint](https://github.com/ChainSafe/mina-rs/commits/noot/checkpoint) continue with issue #24 

- [x] Add `initCheckpoints`, and `isShortRange` algorithms outlined here: https://github.com/MinaProtocol/mina/blob/32f529d8d8d712a44ee75be66061ce08cbdc8924/docs/specs/consensus/README.md#53-decentralized-checkpointing
- [x] Add defined `consensus_state` type.
- [x] Add license in `checkpoint.rs`.
- [x] Update library.
- [x] Fix: Use of moved value.
- [x] Feat: Implement `is_short_range()` for short-range fork check, `test_init_checkpoints` and , `init_checkpoints` for  genesis checkpoints follow the new spec.
- [x] Test: Implement `test_is_short_range()` - Add both true and false test cases.
- [x]  Fix: Replace Common to Chain trait.
- [x]  chore: Recover Cargo.lock.

**Reference issue to close (#24 )**

Closes Closes 
Cargo build & Cargo test passed.
Screenshot (291)
**Specification reference**

- 
https://github.com/MinaProtocol/mina/blob/32f529d8d8d712a44ee75be66061ce08cbdc8924/docs/specs/consensus/README.md

https://github.com/MinaProtocol/mina/blob/develop/src/lib/consensus/proof_of_stake.ml


consensus/src/common.rs
```diff
@@ -30,7 +30,48 @@ where
     fn state_hash(&self) -> Option<StateHash>;
 }
 
-impl Chain<ProtocolState> for ProtocolStateChain {
+impl Common<ProtocolState> for ProtocolState {
```
creativcoder, line 33: This should be reverted back. Can you explain why we had to re-implement this. These methods are common across all ProtocolState instances.

consensus/src/checkpoint.rs
```diff
@@ -0,0 +1,142 @@
...
+        );
+        assert_eq!(
+            genesis.body.consensus_state.next_epoch_data.lock_checkpoint,
+            Base58Encodable::from_base58("3NLoKn22eMnyQ7rxh5pxB6vBA3XhSAhhrf7akdqS6HbAKD14Dh1d")
+                .unwrap()
+        );
+    }
+
+    #[wasm_bindgen_test]
+    fn test_is_short_range() {
```
creativcoder, line 120: This only tests the `true` path. Can we add a test which asserts for alternative test case? Also we can look into proof_of_stake.ml and see if we can include those tests as well.

consensus/src/checkpoint.rs
```diff
@@ -0,0 +1,142 @@
...
+
+use crate::common::{Common, ProtocolStateChain};
+use mina_crypto::base58::Base58Encodable;
+use mina_crypto::hash::{EpochSeed, StateHash};
+use mina_rs_base::protocol_state::ProtocolState;
+use wasm_bindgen_test::*;
+const SLOTS_PER_EPOCH: u32 = 7140;
+
+/// init_checkpoints initializes the checkpoints for the genesis block
+/// This function assumes the state hash of `genesis` is already set
```
creativcoder, line 12: can remove this comment, as we do initialize the two checkpoints within this function.

consensus/src/common.rs
polarbeargo: This is from @noot's legacy code as the following:
Screenshot (285)
Screenshot (286)
We have both ProtocolStateChain and ProtocolState.

polarbeargo: This part is from @noot's legacy code
Screenshot (285)
Screenshot (286)
:

consensus/src/checkpoint.rs
polarbeargo: Will do. I will skip the selection chain test case part because we haven't reached the selection part.

consensus/src/common.rs
creativcoder: As in https://github.com/ChainSafe/mina-rs/pull/34/files#diff-09b1b8c7ec19c8817ec99388da62c7d423b51a868df0ec116ef9d4dde9919286R21 the `Chain` trait should be left as it is. `ProtocolStateChain` only implements that. We would just need to update the methods within that trait to match the new spec, if there are any. If you see this section: https://github.com/MinaProtocol/mina/blob/32f529d8d8d712a44ee75be66061ce08cbdc8924/docs/specs/consensus/README.md#51-common those are all the methods the Chain trait implements. Also we don't need to strictly abide by the exact names. The spec naming convention is just for readability purposes. We can rename variables and type names as it makes sense in our code. So that's why we don't have it as `Common` but rather as `Chain`

polarbeargo: Got it! Thank you for clarifying, Rahul.

consensus/src/common.rs
```diff
@@ -30,6 +30,47 @@ where
     fn state_hash(&self) -> Option<StateHash>;
 }
 
+impl Chain<ProtocolState> for ProtocolState {
```
creativcoder, line 33: I don't see why we need to implement Chain here on ProtocolState, as it's already implemented on `ProtocolStateChain`. Also these API's only make sense on a blockchain. ProtocolState is a just a single block header, so by definition it should not be implementing Chain trait.

polarbeargo: Ok, I will remove `ProtocolState` from  `Chain `trait.

consensus/src/checkpoint.rs
```diff
@@ -0,0 +1,163 @@
...
+            .unwrap()
+            .staking_epoch_data
+            .lock_checkpoint
+            == c1
+                .consensus_state()
+                .unwrap()
+                .next_epoch_data
+                .lock_checkpoint;
+    } else {
+        return false;
```
creativcoder, line 75: we can remove `return` from both branches

polarbeargo: If remove return build will fail.
Screenshot (305)

consensus/src/checkpoint.rs
```diff
@@ -0,0 +1,163 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::common::{Chain, ProtocolStateChain};
+use mina_crypto::base58::Base58Encodable;
+use mina_crypto::hash::{EpochSeed, StateHash};
+use mina_rs_base::protocol_state::ProtocolState;
+#[allow(dead_code)]
```
creativcoder, line 8: I don't think we need this. This is being used below right?

polarbeargo: We use this in the init_checkpoints()

creativcoder: then this attribute should not be there, because the lint says dead_code

polarbeargo: If we don't use this tag these Clippy warnings will come out again

polarbeargo: It actually used at [here ](https://github.com/ChainSafe/mina-rs/blob/a69a2d09029df9ef739a4d4a9e77d8b7b014b39b/consensus/src/checkpoint.rs#L62)but Clippy think it's never used [Github Action](https://github.com/ChainSafe/mina-rs/runs/4284862218?check_suite_focus=true#step:9:20)

creativcoder: ok, let's tackle this later.

creativcoder (approved):

creativcoder:
LGTM, good for merge after a squash

crypto/src/hash/types.rs
```diff
@@ -18,7 +18,7 @@ use wire_type::WireType;
 
 pub(crate) type HashBytes = Box<[u8]>;
 
-#[derive(Clone, Default, Debug, PartialEq, Serialize, Deserialize)]
+#[derive(Copy, Clone, Default, Debug, PartialEq, Serialize, Deserialize)]
```
creativcoder, line 21: Can we not use Copy here. Deriving Copy is implicit here. Instead we can call `clone()` where we need them, or just take ref of the BaseHash

polarbeargo: We have to use Copy to resolve this error "[ Use of moved value.](https://github.com/ChainSafe/mina-rs/pull/80/commits/68772460e890148d6d1630eac2248255bb480ac2)"

consensus/src/checkpoint.rs
```diff
@@ -0,0 +1,163 @@
...
+        .lock_checkpoint = StateHash::default();
+    genesis
+        .body
+        .consensus_state
+        .staking_epoch_data
+        .epoch_length
+        .0 = 1;
+    genesis.body.consensus_state.next_epoch_data.seed =
+        Base58Encodable::from_base58("2vaRh7FQ5wSzmpFReF9gcRKjv48CcJvHs25aqb3SSZiPgHQBy5Dt")
+            .unwrap();
```
hanabi1224, line 32: We should not call `.unwrap` in non-test code, either return Result or use expect

consensus/src/checkpoint.rs
```diff
@@ -0,0 +1,163 @@
...
+    genesis
+        .body
+        .consensus_state
+        .next_epoch_data
+        .start_checkpoint = StateHash::default();
+    genesis.body.consensus_state.next_epoch_data.lock_checkpoint =
+        Base58Encodable::from_base58("3NLoKn22eMnyQ7rxh5pxB6vBA3XhSAhhrf7akdqS6HbAKD14Dh1d")
+            .unwrap();
```
hanabi1224, line 40: ditto

consensus/src/checkpoint.rs
```diff
@@ -0,0 +1,163 @@
...
+        .staking_epoch_data
+        .lock_checkpoint = StateHash::default();
+    genesis
+        .body
+        .consensus_state
+        .staking_epoch_data
+        .epoch_length
+        .0 = 1;
+    genesis.body.consensus_state.next_epoch_data.seed =
+        Base58Encodable::from_base58("2vaRh7FQ5wSzmpFReF9gcRKjv48CcJvHs25aqb3SSZiPgHQBy5Dt")
```
hanabi1224, line 31: if this is a static array, use &'static T in the struct instead

consensus/src/checkpoint.rs
```diff
@@ -0,0 +1,163 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::common::{Chain, ProtocolStateChain};
+use mina_crypto::base58::Base58Encodable;
+use mina_crypto::hash::{EpochSeed, StateHash};
+use mina_rs_base::protocol_state::ProtocolState;
+#[allow(dead_code)]
+const SLOTS_PER_EPOCH: u32 = 7140;
+
+#[allow(dead_code)]
```
hanabi1224, line 11: why dead code instead of making the funs pub?

consensus/src/checkpoint.rs
```diff
@@ -0,0 +1,163 @@
...
+            .staking_epoch_data
+            .lock_checkpoint
+            == c1
+                .consensus_state()
+                .unwrap()
+                .staking_epoch_data
+                .lock_checkpoint;
+    }
+
+    #[allow(clippy::erasing_op)]
```
hanabi1224, line 57: why is this needed?

consensus/src/checkpoint.rs
```diff
@@ -0,0 +1,161 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::common::{Chain, ProtocolStateChain};
+use mina_crypto::base58::Base58Encodable;
+use mina_crypto::hash::{EpochSeed, StateHash};
+use mina_rs_base::protocol_state::ProtocolState;
+
+pub const SLOTS_PER_EPOCH: u32 = 7140;
+
+pub fn init_checkpoints(genesis: &mut ProtocolState) {
```
hanabi1224, line 11: why not impl init_checkpoints  and is_short_range  for ProtocolState and ProtocolStateChain respectively

polarbeargo: They're not pure functions as params are mutable

consensus/src/lib.rs
```diff
@@ -1,10 +1,5 @@
 // Copyright 2020 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0
 
-#![deny(rustdoc::all)]
```
hanabi1224, line 4: plz revert this bad merge

consensus/src/checkpoint.rs
```diff
@@ -0,0 +1,158 @@
...
+        Base58Encodable::from_base58("3NLoKn22eMnyQ7rxh5pxB6vBA3XhSAhhrf7akdqS6HbAKD14Dh1d")
+            .map_err(|_| ConsensusErrTyp::ConsensusInitFail)?;
+    genesis.body.consensus_state.next_epoch_data.epoch_length.0 = 2;
+    Ok(())
+}
+
+pub fn is_short_range(
+    c0: &ProtocolStateChain,
+    c1: &ProtocolStateChain,
+) -> Result<bool, ConsensusErrTyp> {
```
hanabi1224, line 51: OK for now, but if u look at [ocaml code](https://github.com/MinaProtocol/mina/blob/9dd0c963e87f80c81006c6ecda34efbf6543d2a3/src/lib/consensus/proof_of_stake.ml#L2919), I think we can safely return false for the err condition

polarbeargo: For not use unwrap() https://github.com/ChainSafe/mina-rs/pull/80/commits/c15385f425f4344052ef57f3e27bcee1a5f621c2#

consensus/src/checkpoint.rs
```diff
@@ -0,0 +1,159 @@
...
+    let s0_lock_checkpoint = &s0.staking_epoch_data.lock_checkpoint;
+    let s1_lock_checkpoint = &s1.staking_epoch_data.lock_checkpoint;
+    let s1_next_epoch_lock_checkpoint = &s1.next_epoch_data.lock_checkpoint;
+
+    if s0.epoch_count == s1.epoch_count {
+        return Ok(s0_lock_checkpoint == s1_lock_checkpoint)
+    }
+
+    if s0.epoch_count.0 == s1.epoch_count.0 + 1
+        && Chain::epoch_slot(c1) >= Some(SLOTS_PER_EPOCH * 2 / 3)
```
creativcoder, line 67: instead of `Chain::epoch_slot`, we can do `c1.epoch_slot()`

hanabi1224 (approved):

consensus/src/checkpoint.rs
polarbeargo: Thank you @creativcoder 😄

PR #79, opened 2021-11-09
Refine continuous integration setup

hanabi1224:
**Summary of changes**
Changes introduced in this pull request:
- Merge 'check license' with 'cargo fmt check' into checks
- Remove stale working-bee branch in the setup
- Group jobs into checks, x86_64, wasm and codecov-x86_64
- Avoid allocating CI VM for every small job (although GH does not have rate limit atm)
- Run wasm tests for test-serialization against both node and headless chrome context
- Add dependency badge
- Replace/Remove insecure dependencies shown here: https://deps.rs/repo/github/ChainSafe/mina-rs
- Add [`cargo deny check`](https://embarkstudios.github.io/cargo-deny/) to `checks` job
- Add benchmark tests for block(ExternalTransition) serde
- Add `cargo bench` to x86_64 job
- Upload benchmark (html) report as build artifacts

**Reference issue to close (if applicable)**

Closes 

**Other information and links**

PR #81, opened 2021-11-11
Refactor protocol state proof

willemolding:
**Summary of changes**
- Refactors `ProtocolStateProof` in preparation for verification
- Renames some types to better match those in arkworks and o(1)/proof-systems
    - In particular makes it clear which fields are finite field/ elliptic curve points
- Refactors `protocol_state_proof.rs` into a number of files for readability
- Adds conversions to arkworks and proof-systems types where the conversion is obvious

- https://github.com/o1-labs/proof-systems/blob/master/dlog/plonk/src/prover.rs


base/Cargo.toml
```diff
@@ -15,4 +15,11 @@ chrono = "0.4"
 derive_deref = "1.1"
 num = "0.4"
 serde = { version = "1", features = ["derive"] }
-thiserror = "1"
\ No newline at end of file
+thiserror = "1"
+
+plonk_protocol_dlog = { git = "https://github.com/o1-labs/proof-systems" }
```
hanabi1224, line 20: need to whitelist 'o1-labs' in deny.toml after #79 is merged

willemolding: Good to know

base/src/protocol_state_proof/opening_proof.rs
```diff
@@ -0,0 +1,37 @@
...
+    pub sg: FiniteECPoint,
+}
+
+impl<P> From<OpeningProof> for commitment_dlog::commitment::OpeningProof<GroupAffine<P>>
+where
+    P: ark_ec::SWModelParameters,
+    <P as ModelParameters>::BaseField: From<ark_ff::BigInteger256>,
+    <P as ModelParameters>::ScalarField: From<ark_ff::BigInteger256>,
+{
+    fn from(t: OpeningProof) -> Self {
```
hanabi1224, line 28: Just to note, there seems to be no code coverage for all newly added from trait impls, we need to improve that if not applicable in this PR.

https://app.codecov.io/gh/ChainSafe/mina-rs/compare/81/diff#D4L28

hanabi1224 (approved):

base/Cargo.toml
```diff
@@ -15,4 +15,11 @@ chrono = "0.4"
...
+plonk_circuits = { git = "https://github.com/o1-labs/proof-systems" }
```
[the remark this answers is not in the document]
creativcoder: Yes, definitely

base/src/numbers.rs
```diff
@@ -123,10 +123,16 @@ impl BlockTime {
...
 pub struct BlockTimeSpan(pub u64);
 
-// Consider switch to [ark-ff](https://docs.rs/ark-ff/0.3.0/ark_ff/biginteger/struct.BigInteger256.html)
 #[derive(Clone, Serialize, Deserialize, Default, PartialEq, Debug)]
 pub struct BigInt256(pub [u8; 32]);
 
+impl From<BigInt256> for ark_ff::BigInteger256 {
+    fn from(i: BigInt256) -> Self {
+        use ark_ff::bytes::FromBytes;
+        Self::read(&i.0[..]).unwrap()
```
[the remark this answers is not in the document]
creativcoder: I think implementing Deref it will expose internal details (if we want to hide it in future) and was only intended to implement for smart pointer types when it got introduced. More details: https://stackoverflow.com/questions/45086595/is-it-considered-a-bad-practice-to-implement-deref-for-newtypes

Again it depends if we not want to expose the internal impl of the newtype.

willemolding:
I have added the MPL-2.0 license to the allowable list based on reading this description https://fossa.com/blog/open-source-software-licenses-101-mozilla-public-license-2-0/

In summary it is a weak copyleft so we can use it in derived works under a different license as long as we don't change the original source code (which we are not)

base/src/protocol_state_proof/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+    fn default() -> Self {
+        Self::ShiftedValue(BigInt256::default())
+    }
+}
+
+#[derive(Clone, Serialize, Deserialize, Default, PartialEq, Debug, WireType)]
+#[serde(from = "<Self as WireType>::WireType")]
+#[serde(into = "<Self as WireType>::WireType")]
+#[wire_type(recurse = 2)]
+pub struct SpongeDigestBeforeEvaluations((Hex64, Hex64, Hex64, Hex64, ()));
```
creativcoder, line 100: is the empty unit needed for a byte in binprot or is it something from arkworks?

creativcoder (approved):
Yay, great start towards m3!

base/src/protocol_state_proof/mod.rs
willemolding: It is for bin-prot. They use null byte terminated arrays and this is a nice trick for deserializing those with serde

Issue #66, opened 2021-10-26
M1 - Add the genesis block


creativcoder:
**Issue summary**

We now are able to read blocks from the network. It would be great to have the genesis block information implemented as part of M1 where the node would initialize it from the db. We can keep the db in-memory till we create a crate for abstracting the Db impl.

AC:

* Add an `initialize_genesis` API
* Load genesis block
* Test serialization and deserialization of genesis block from mainnet.

**Other information and links**

mina block explorer: https://graphql.minaexplorer.com/

graphql query for block at height 1 (genesis block): 
```graphql
{
  blocks(query: { canonical: true }, sortBy: DATETIME_ASC, limit: 1) {
    creatorAccount {
      publicKey
    }
    blockHeight
    stateHash
    dateTime
    protocolState {
      consensusState {
        blockHeight
      }
    }
    transactions {
      coinbase
    }
  }

```

PR #90, opened 2021-11-29
[M2] Genesis block initialization

hanabi1224:
**Summary of changes**
- spec: https://github.com/MinaProtocol/mina/tree/feature/9665-spec-ouroboros-samasika-checkpointing/docs/specs/consensus#611-genesis-block
- genesis block initialization
- [M1][M2] genesis block full parity test (mainnet only)

Closes 
https://github.com/ChainSafe/mina-rs/issues/66


consensus/src/genesis/genesis_init_config.rs
```diff
@@ -0,0 +1,19 @@
...
+
+lazy_static::lazy_static! {
+    pub static ref MAINNET_CONFIG: GenesisInitConfig = GenesisInitConfig::mainnet();
+    pub static ref DEVNET_CONFIG: GenesisInitConfig = GenesisInitConfig::devnet();
+}
+
+pub struct GenesisInitConfig {}
+
+impl GenesisInitConfig {
+    pub(crate) fn mainnet() -> Self {
```
creativcoder, line 12: Do you think we can instead have two modules like:
```
consensus/config/mainet.rs
consensus/config/devnet.rs
```
`with #![cfg(feature = "<network>")]` in them? where `network = devnnet | mainnet`

Reason: I believe quite a few areas would need separate but similar functions for devnet and mainnet and that they would grow in future, so instead of having both APIs for devnet and mainnet together in the compiled library, we could just include either of them based on feature flag.

Also maybe in future we could extract config module to one level up so that it would contain config for other components as well, such as network configs and others.

consensus/src/genesis/genesis_init.rs
```diff
@@ -0,0 +1,24 @@
...
+///
+/// # Example
+/// ```rust
+/// use mina_rs_base::types::*;
+/// use mina_consensus::genesis::{GenesisInit, DEVNET_CONFIG, MAINNET_CONFIG};
+/// let genesis_mainnet = ExternalTransition::init_genesis(&MAINNET_CONFIG);
+/// let genesis_devnet = ExternalTransition::init_genesis(&DEVNET_CONFIG);
+/// ```
+pub trait GenesisInit {
+    fn init_genesis(config: &GenesisInitConfig) -> ExternalTransition;
```
creativcoder, line 17: I wonder if we could do something like:

```
trait GenesisInit {
    const CONFIG: Config = config::load()
}
```
config::load would load from conditionally compiled modules.
what do you think?

consensus/src/genesis/genesis_init_config.rs
hanabi1224: We can use conditional compilation in bin crate, making devnet | mainnet config/feature orthogonal in the lib crate might not be a good idea for some cases, e.g. testing, we will have to configure different feature set for the tests to cover all scenarios

consensus/src/genesis/genesis_init.rs
hanabi1224: That will be great but I doubt everything can fit into const fn (e.g. PublicKeys)

consensus/src/genesis/genesis_init_config.rs
creativcoder: yeah, it make sense at the binary level.

consensus/src/genesis/genesis_init.rs
creativcoder: right

protocol/test-serialization/src/genesis.rs
```diff
@@ -0,0 +1,272 @@
...
+                "d96d62e54a0a49d3a44c919eb4b089333d64a236edcda1921274ac6903bad937"
+            );
+            assert_eq!(ev0.t.0.len(), 5);
+        }
+    }
+
+    #[test]
+    #[wasm_bindgen_test]
+    fn test_genesis_consensus_mainnet() {
+        test_genesis_consensus(&GENESIS_BLOCK_MAINNET, &MAINNET_CONFIG)
```
hanabi1224, line 65: parity test 1

protocol/test-serialization/src/genesis.rs
```diff
@@ -0,0 +1,272 @@
...
+        genesis_init_config: &GenesisInitConfig,
+    ) {
+        let genesis = ExternalTransition::init_genesis(genesis_init_config);
+        let output = genesis.try_serialize().unwrap();
+        assert_eq!(genesis_fixture.bytes, output)
+    }
+
+    #[test]
+    #[wasm_bindgen_test]
+    fn test_genesis_path_mainnet() {
```
hanabi1224, line 85: parity test 2

protocol/test-serialization/src/test_3NKaBJsN1SehD6iJwRwJSFmVzJg5DXSUQVgnMxtH4eer4aF5BrDK.rs
```diff
@@ -319,14 +302,20 @@ mod tests {
...
 
-        let hash_str = "jwHLk8kaC6B45K3sjuX2sM38649VtfpUAteTfKFQMPcqTeXjGiT";
-        let bytes = bs58::decode(hash_str).into_vec().unwrap();
-        assert_eq!(et.delta_transition_chain_proof.0.as_ref(), &bytes[2..34]);
-        // FIXME: Investigate why this fails
+        let bytes = bs58::decode("jwHLk8kaC6B45K3sjuX2sM38649VtfpUAteTfKFQMPcqTeXjGiT")
+            .into_vec()
+            .unwrap();
+        assert_eq!(et.delta_transition_chain_proof.0.as_ref()[..], bytes[2..34]);
+        // FIXME:
```
creativcoder, line 309: next PR or missed?

protocol/bin-prot/src/bin_prot_deser.rs
```diff
@@ -0,0 +1,26 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::de::Deserializer;
+use serde::{Deserialize, Serialize};
+
+pub trait BinProtDeser {
```
creativcoder, line 7: It's not clear as to why we need to create a trait for the methods below? Are tests the only place those methods are used, if yes, I think we can have those methods in tests itself rather than being part of the library. Thoughts?

crypto/src/hash/types.rs
```diff
@@ -166,7 +143,12 @@ pub struct NonSnarkStagedLedgerHash {
 #[derive(Clone, Default, Debug, PartialEq, Serialize, Deserialize, WireType)]
 #[serde(from = "<Self as WireType>::WireType")]
 #[serde(into = "<Self as WireType>::WireType")]
-pub struct AuxHash(Vec<u8>);
+pub struct AuxHash(pub Vec<u8>);
+
+// FIXME:
```
creativcoder, line 148: next PR or missed?

crypto/src/base58/encodable.rs
```diff
@@ -20,17 +18,105 @@ pub trait Base58Encodable {
...
+                message: format!("{:?}", e),
+            })?;
 
         // skip the first byte as this still contains the version byte
-        Ok(from_reader(&bytes[1..]).unwrap())
+        from_reader(&bytes[1..])
+    }
+}
+
+pub trait Base58EncodableHash {
```
creativcoder, line 48: is it possible to combine `Base58Encodable` and `Base58EncodableHash`? both seems to overlap in terms of their API

crypto/src/base58/encodable.rs
```diff
@@ -20,17 +18,105 @@ pub trait Base58Encodable {
...
+    /// This is the only part a custom implementation need provide.
+    /// Should be a constant from the base58_version_bytes.rs file corresponding
+    /// to the type.
+    const VERSION_BYTE: u8;
+    const MINA_VERSION_BYTE_COUNT: usize = 1;
+    const MINA_VERSION_BYTE: u8 = 1;
+    fn to_base58(&self) -> EncodeBuilder<'static, Vec<u8>>
```
creativcoder, line 55: The return type currently restricts `&self` to `'static`. was it not possible with any `'a`?

base/src/numbers.rs
```diff
@@ -25,6 +25,12 @@ use wire_type::WireType;
 #[wire_type(recurse = 2)]
 pub struct Length(pub u32);
 
+impl From<u32> for Length {
```
creativcoder, line 28: we can use [derive_more](https://github.com/JelteF/derive_more) if we want

base/src/consensus_state.rs
```diff
@@ -14,7 +14,17 @@ use wire_type::WireType;
 #[derive(Clone, Serialize, Deserialize, Default, PartialEq, Debug, WireType)]
 #[serde(from = "<Self as WireType>::WireType")]
 #[serde(into = "<Self as WireType>::WireType")]
-pub struct VrfOutputTruncated(Vec<u8>);
+pub struct VrfOutputTruncated(pub Vec<u8>);
+
+impl VrfOutputTruncated {
```
willemolding, line 19: What's this about Base64, I thought that only Base58 was used in Mina. If both are used we should add a new trait `Base64Encodable` that mirrors `Base58Encodable` and implement that here

base/src/protocol_state_proof/mod.rs
```diff
@@ -142,3 +142,11 @@ pub struct ProofOpenings {
     pub proof: OpeningProof,
     pub evals: (ProofEvaluations, ProofEvaluations),
 }
+
+pub trait HexConvertable {
```
willemolding, line 146: This feels like a strange place for this trait to live. Maybe we can group all the Base58, Base64, Hex conversion traits in the same place

consensus/src/genesis/config.rs
```diff
@@ -0,0 +1,542 @@
...
+    base58::{version_bytes, Base58Encodable, Base58EncodableHash},
+    hash::*,
+    signature::PublicKey,
+};
+use mina_rs_base::{
+    finite_ec_point, finite_ec_point_pair,
+    types::{proof_messages::ProofMessageWithoutDegreeBoundList, *},
+};
+
+const ERR_FAIL_TO_DECODE_B58: &str = "Failed to decode hash from base58";
```
willemolding, line 14: This doesn't feel idiomatic to me. I guess it's ok since this for the given values we can be sure this function can't error. Maybe add some comments justifying why this is how you have done it this way so a new dev doesn't copy this error handling pattern somewhere else :P

crypto/src/base58/encodable.rs
```diff
@@ -20,17 +18,105 @@ pub trait Base58Encodable {
...
+    fn from_base58<'a, I>(i: I) -> Result<Self, bin_prot::error::Error>
     where
         I: AsRef<[u8]>,
         Self: Sized + Deserialize<'a>,
     {
-        let bytes: Vec<u8> = decode(i).with_check(Some(Self::VERSION_BYTE)).into_vec()?;
+        let bytes: Vec<u8> = bs58::decode(i)
+            .with_check(Some(Self::VERSION_BYTE))
+            .into_vec()
+            .map_err(|e| bin_prot::error::Error::Custom {
```
willemolding, line 39: I'm not a fan of this use of bin_prot::Error.

Define a new error type for this trait if needed using `thiserror` it much more straightforward and readable

crypto/src/base58/encodable.rs
```diff
@@ -20,17 +18,105 @@ pub trait Base58Encodable {
...
+
+    fn from_base58<I>(i: I) -> Result<Self, bin_prot::error::Error>
+    where
+        I: AsRef<[u8]>,
+        Self: From<[u8; 32]>,
+    {
+        let bytes: Vec<u8> = bs58::decode(i)
+            .with_check(Some(Self::VERSION_BYTE))
+            .into_vec()
+            .map_err(|e| bin_prot::error::Error::Custom {
```
willemolding, line 84: same applies here

crypto/src/base58/encodable.rs
willemolding: I'm a bit confused why this trait is required at all. Lets catch up to talk about it

crypto/src/base58/encodable.rs
```diff
@@ -20,17 +18,105 @@ pub trait Base58Encodable {
...
+
+        // skip the bs58 version byte and mina bin_prot version bytes
+        let mut b32 = [0; 32];
+        b32.copy_from_slice(&bytes[(1 + Self::MINA_VERSION_BYTE_COUNT)..]);
+        Ok(b32.into())
+    }
+}
+
+#[macro_export]
+macro_rules! impl_hash_bs58_full {
```
willemolding, line 96: This is cool but I think a more idiomatic approach would be to make a custom derive macro

crypto/src/signature.rs
```diff
@@ -52,6 +59,47 @@ impl Signature {
     pub fn inner_curve_scalar(&self) -> &InnerCurveScalar {
         &self.0 .1
     }
+
+    pub fn to_base58(&self) -> EncodeBuilder<'static, Vec<u8>> {
```
willemolding, line 63: This should be in the implementation of `Base58Encodeable`. You can add a custom implementation of to_base58 from that trait no need to use the default one

protocol/bin-prot/src/bin_prot_deser.rs
willemolding: I see what you are trying to do here but I agree we should keep it out of the public API for `bin_prot` and try and keep it to the standard serde api where possible

willemolding (changes requested):

crypto/src/base58/encodable.rs
willemolding: actually having thought about this it is kind of nicer this way. @creativcoder do you have any thoughts on this?

willemolding (changes requested):

base/src/numbers.rs
hanabi1224: Good suggestion, fixed!

base/src/consensus_state.rs
hanabi1224: fixed

crypto/src/base58/encodable.rs
hanabi1224: Combined with sacrificing some performance

crypto/src/hash/types.rs
hanabi1224: fixed

crypto/src/base58/encodable.rs
hanabi1224: That's just a delegate of the type of `base58::encode`

base/src/protocol_state_proof/mod.rs
hanabi1224: Fixed

protocol/bin-prot/src/bin_prot_deser.rs
hanabi1224: Fixed, renamed to BinProtEncodable under crypto

consensus/src/genesis/config.rs
hanabi1224: Justification added

protocol/test-serialization/src/test_3NKaBJsN1SehD6iJwRwJSFmVzJg5DXSUQVgnMxtH4eer4aF5BrDK.rs
hanabi1224: the hex representation from block json uses a different byte code here, which is unexpected, need to figure out why

crypto/src/base58/encodable.rs
hanabi1224: Fixed

crypto/src/base58/encodable.rs
hanabi1224: I believe it's preferable to do it this way unless the procedural macro is meant to part of public APIs, e.g.  https://github.com/rust-lang/portable-simd/blob/master/crates/core_simd/src/vector/float.rs#L208

protocol/bin-prot/src/bin_prot_deser.rs
```diff
@@ -0,0 +1,26 @@
...
+    const PREALLOCATE_BUFFER_BYTES: usize;
+
+    fn try_serialize(&self) -> Result<Vec<u8>, crate::error::Error>
```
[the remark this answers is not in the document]
hanabi1224: I didn't meant to make this a general trait for serialization / deserialization, but specific to bin_prot format actually, plz take a look at the refactored code and let me know your feedback, thx

crypto/src/base58/encodable.rs
creativcoder: Yeah, I think if it simplifies the repetitive parts of code without sacrificing readability of the macro impl (which i think is simple enough here), we're fine with this.

crypto/src/signature.rs
hanabi1224: Fixed.

consensus/src/genesis/genesis_init.rs
```diff
@@ -0,0 +1,187 @@
...
+}
+
+impl GenesisInit for ExternalTransition {
```
creativcoder, line 19: minor, but i think we can call this trait simply `Genesis` and have a method name `from_genesis` instead of `init_genesis`, as `from_*` is the usual method naming convention in Rust codebase, when deriving a value out of some other value. Thoughts?

hanabi1224: How about from_genesis_config ?

creativcoder: yeah

hanabi1224: fixed

creativcoder (approved):
great start towards M2!

polarbeargo (approved):

polarbeargo (approved):
Nice work very elegant 👍 !!!

base/src/consensus_state.rs
polarbeargo: Nice job 💯

base/src/numbers.rs
```diff
@@ -175,33 +154,38 @@ pub struct BlockTimeSpan(pub u64);
...
+    fn to_hex_string(&self) -> String
+    where
+        Self: AsRef<[u8]>,
+    {
+        hex::encode(self)
+    }
+
+    fn try_from_hex(s: impl AsRef<[u8]>) -> Result<Self, Self::Error> {
+        let mut s = s.as_ref();
+        if s[1] == b'x' && (s[0] == b'0' || s[0] == b'\\') {
```
willemolding, line 175: I think a refactor would be possible here. I doesn't seem right to be to have the parsing of strings (which would be the same for all hex strings) in the `impl` specific to `BigInt256`

base/src/protocol_state_proof/field_and_curve_elements.rs
```diff
@@ -51,7 +31,7 @@ impl HexConvertable for FieldElementVec {
         s
     }
 
-    fn try_from_hex_string(s: impl AsRef<[u8]>) -> Result<Self, Self::Error> {
+    fn try_from_hex(s: impl AsRef<[u8]>) -> Result<Self, Self::Error> {
```
willemolding, line 34: same here

crypto/src/binprot/mod.rs
```diff
@@ -1,22 +1,23 @@
 // Copyright 2020 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::de::Deserializer;
+pub use bin_prot::error::Error;
+use bin_prot::Deserializer;
 use serde::{Deserialize, Serialize};
 
-pub trait BinProtDeser {
+pub trait BinProtEncodable {
```
willemolding, line 8: Ok I understand why this needs to be BinProt specific now, since then you can know exactly how much buffer to allocate

crypto/src/base58/encodable.rs
```diff
@@ -1,122 +1,108 @@
...
+            fn try_from(h: Vec<u8>) -> Result<Self, Self::Error> {
+                bin_prot::from_reader(h.as_slice())
+            }
+        }
+    };
 }
 
 #[macro_export]
-macro_rules! impl_hash_bs58_full {
+macro_rules! impl_bs58_for_hash_full {
```
willemolding, line 76: This is kind of a nitpick but this macro isn't just for hashes, it is for any strings of bytes right? This is a recurring source of confusion. To me a hash specifically means 'the output of a hash function'

crypto/src/lib.rs
```diff
@@ -16,5 +16,17 @@
 // attribute macro and an order of operations well defined
 
 pub mod base58;
+pub mod base64;
+pub mod binprot;
 pub mod hash;
+pub mod hex;
 pub mod signature;
+
+pub mod prelude {
```
willemolding, line 25: I think it is best practice to only export traits in a prelude so consider removing the hash and signature types?

crypto/src/base58/encodable.rs
```diff
@@ -1,122 +1,108 @@
...
-    /// Should be a constant from the base58_version_bytes.rs file corresponding
-    /// to the type.
-    const VERSION_BYTE: u8;
     const MINA_VERSION_BYTE_COUNT: usize = 1;
     const MINA_VERSION_BYTE: u8 = 1;
-    fn to_base58(&self) -> EncodeBuilder<'static, Vec<u8>>
-    where
-        Self: AsRef<[u8; 32]>,
-    {
+    fn to_base58(&self) -> EncodeBuilder<'static, Vec<u8>> {
```
willemolding, line 14: I feel strongly that `Base58Encodable` and `Base64Encodable` should have an identical interface since they only differ in the encoding of the output not how the byte representation is accessed

willemolding (changes requested):
Still more changes requested! Sorry :stuck_out_tongue:

crypto/src/base58/encodable.rs
hanabi1224: fixed

crypto/src/base58/encodable.rs
hanabi1224: In that case, the 2 traits are kind of exclusive to each other, ppl will encounter fn naming conflict if they try to use both traits in the same file, and it will it hard for ppl who read the code later to know exactly which encoding it's doing without looking up in the context

crypto/src/lib.rs
hanabi1224: fixed

base/src/numbers.rs
hanabi1224: Sry but did not quite get ur point, this is implementing HexEncodable trait for BigInt256 not for the struct itself, or do you suggest not implementing the trait? Could you plz elaborate on what would be a better design here?

crypto/src/base58/encodable.rs
willemolding: Sorry my comment didn't really communicate what I meant. They should use the same process for accessing the underlying bytes. It feels strange for me for `Base64Encodable` to require an `AsRef` implementation to get the bytes and for `Base58Encodable` to require implementing a function to get the bytes. Either is fine but I think they should be the same.

Definitely the public trait interface should be different e.g. `to_base58` and `to_base64`

willemolding: Hmm the `AsRef<[u8]>` approach only works if the structure can provide the bytes as slice of contiguous memory it owns. Maybe the `get_bytes` approach is more general

hanabi1224: The core difference is that, for mina base58 encoding impls, we have 2 different scenarios that the struct may or may not own the contiguous memory, for either case, we need to deal with the extra mina version bytes hence there has to be owned Vec<u8> as intermediate mem holder.   But for mina base64 encoding case, the owned intermediate mem holder is not needed in any cases, it will just add unnecessary overhead(heap allocation) to create an owned intermediate mem holder if we want to unify the way how to get the encodable bytes.  I've renamed the trait fn in Base58Encodable to make it more clear.  Please note that these 2 encodable trait is only designed to be mina specific and I don't feel that we really want to unify the API at the cost of  extra performance overhead, does that make sense?

willemolding: It makes sense but still doesn't feel right to me. In the interest of getting this PR moving I'll accept what we have and we can maybe revisit it again later.

base/src/numbers.rs
willemolding: In both impls of `try_from_hex` you copy the code

```rust
if s[1] == b'x' && (s[0] == b'0' || s[0] == b'\\') {
    s = &s[2..];
}
```
It would be nice not to have this repeated.

I think it would make more sense to assume that the input slice only contains the actual hex of the data structure (not string encoded with a `0x` prefix) and handle stripping the prefix in the calling code. Possible refactor the prefix stripping to a function

willemolding (changes requested):

base/src/numbers.rs
hanabi1224: > Possible refactor the prefix stripping to a function

Fixed

> handle stripping the prefix in the calling code.

There're tons of occurrence for the hardcoded values with either 0x prefix (in json) or \x prefix (in ocaml code),  I don't feel it's worth the effort to manually strip those prefixes during copy & pasting, and we will need to do all those again when we define genesis config for devnet, does that make sense?

creativcoder:
Gonna merge this and take up the remaining changes in a subsequent PR to avoid the overhead of resolving merge conflicts while we're waiting.

creativcoder (approved):

PR #93, opened 2021-12-03
[M2] Test cases for short range fork

### LINKED ISSUE #72: Test coverage for WebAssembly target on CI
**Issue summary**

https://rustwasm.github.io/docs/wasm-bindgen/wasm-bindgen-test/usage.html

This will also require a test harness to be setup for browser environment: https://rustwasm.github.io/docs/wasm-bindgen/wasm-bindgen-test/browsers.html

a/c:

polarbeargo:
**Summary of changes**
- 
Going through the OCaml codebase to create our own test cases manually no need to be the same as Mina Ocaml.

- [x] test: Implement gen_spot_pair_common_checkpoints() reference from Mina Ocaml [line 4111](https://github.com/MinaProtocol/mina/blob/af76cb7980d5e81e704120290a850ea9c6f8522e/src/lib/consensus/proof_of_stake.ml#L4111)
- [x] test: Implement gen_num_blocks_in_epochs() reference from Mina Ocaml [line 3881](https://github.com/MinaProtocol/mina/blob/af76cb7980d5e81e704120290a850ea9c6f8522e/src/lib/consensus/proof_of_stake.ml#L3881)
- [x]  test: Implement gen_spot_root_epoch_position()  reference from Mina Ocaml [line 3960](https://github.com/MinaProtocol/mina/blob/af76cb7980d5e81e704120290a850ea9c6f8522e/src/lib/consensus/proof_of_stake.ml#L3960)
- [x]  test: Implement gen_num_blocks_in_slots() reference from Mina Ocaml [line 3867](https://github.com/MinaProtocol/mina/blob/af76cb7980d5e81e704120290a850ea9c6f8522e/src/lib/consensus/proof_of_stake.ml#L3867)
- [x] test: implement equal_state_in_short_fork_range() reference from Mina Ocaml [line 4394](https://github.com/MinaProtocol/mina/blob/af76cb7980d5e81e704120290a850ea9c6f8522e/src/lib/consensus/proof_of_stake.ml#L4394)
- [x] test: Implement gen_spot_pair_short_aligned_generates_pairs_of_states_in_short_fork_range() reference  from Mina Ocaml [line 4258](https://github.com/MinaProtocol/mina/blob/af76cb7980d5e81e704120290a850ea9c6f8522e/src/lib/consensus/proof_of_stake.ml#L4258)
- [x] test: Implement gen_spot_pair_short_misaligned_generates_pairs_of_states_in_short_fork_range() reference from Mina Ocaml [line 4266](https://github.com/MinaProtocol/mina/blob/af76cb7980d5e81e704120290a850ea9c6f8522e/src/lib/consensus/proof_of_stake.ml#L4266)
- [x] test: Implement gen_spot_pair_long_generates_pairs_of_states_in_long_fork_range() reference from Mina Ocaml [line 4291](https://github.com/MinaProtocol/mina/blob/af76cb7980d5e81e704120290a850ea9c6f8522e/src/lib/consensus/proof_of_stake.ml#L4291)
- [x] Test coverage for WebAssembly target on CI #72

**Reference issue to close ( if applicable )**

Closes 

**Specification reference**

https://github.com/MinaProtocol/mina/blob/develop/src/lib/consensus/proof_of_stake.ml


base/src/consensus_state.rs
```diff
@@ -70,6 +73,91 @@ pub struct ConsensusState {
 }
 
 impl ConsensusState {
+    // TODO: read from config
+    pub fn new() -> Self {
```
hanabi1224, line 77: better to impl Default trait when it has hard coded values

consensus/src/checkpoint.rs
```diff
@@ -112,47 +117,221 @@ mod tests {
...
+        //  staking epoch starts at (the simulation of all blocks preceeding the
+        //  staking epoch
+        let root_epoch_int = thread_rng().gen_range(0..100);
+        let root_block_height =
+            gen_num_blocks_in_epochs(slot_fill_rate, slot_fill_rate_delta, root_epoch_int as f64);
+        return (root_epoch_int, root_block_height as u32);
+    }
+
+    fn convert(x: f64) -> i32 {
+        x.round().rem_euclid(2f64.powi(32)) as u32 as i32
```
hanabi1224, line 143: Perf: Define constant for `2f64.powi(32)`

consensus/src/checkpoint.rs
```diff
@@ -112,47 +117,221 @@ mod tests {
...
+    fn gen_spot(block: &mut ProtocolState) {
+        // New default consensus state
+        // Generate blockchain position and epoch lengths.
+        // staking_epoch == root_epoch, next_staking_epoch == root_epoch + 1
+        // New state slot and length.
+        // New total currency for state.
+        // New epoch data for staking and next epochs.
+        // New block reward information (unused in chain selection).
+        let consensus_state = ConsensusState::new();
+        block.body.consensus_state = consensus_state;
```
hanabi1224, line 163: nit: make this one-liner?

consensus/src/checkpoint.rs
```diff
@@ -112,47 +117,221 @@ mod tests {
...
+        a: &mut ProtocolState,
+        b: &mut ProtocolState,
+        min_a_curr_epoch_slot: u32,
+    ) {
+        // New pairs of spot blocks that share common checkpoints.
+        // The overlap of the checkpoints and the root epoch positions of the blocks
+        // that are generated can be configured independently so that this function
+        // can be used in other generators that wish to generates pairs of spot blocks
+        // with specific constraints.
+        let default_slot_fill_rate = 0.65;
```
hanabi1224, line 176: define these as const

consensus/src/checkpoint.rs
```diff
@@ -112,47 +117,221 @@ mod tests {
...
+        let default_slot_fill_rate_delta = 0.15;
```
hanabi1224, line 177: ditto

consensus/src/checkpoint.rs
```diff
@@ -112,47 +117,221 @@ mod tests {
...
+        let max_epoch_slot = match blockchain_length_relativity {
+            Some(blockchain_length_relativity) => {
+                a.body.consensus_state.curr_global_slot.slots_per_epoch.0 - 4
+            } // -1 to bring into inclusive range, -3 to provide 2 slots of fudge room
+            None => a.body.consensus_state.curr_global_slot.slots_per_epoch.0 - 1, // -1 to bring into inclusive range
+        };
+
+        let min_a_curr_epoch_slot_defaut = 0;
+        let min_a_curr_epoch_slot_sum = min_a_curr_epoch_slot_defaut + min_a_curr_epoch_slot;
+        let slot = thread_rng().gen_range(min_a_curr_epoch_slot_sum..max_epoch_slot);
```
hanabi1224, line 194: U can generate f64 number directly, see https://docs.rs/rand/latest/rand/trait.Rng.html#example-1

consensus/src/checkpoint.rs
```diff
@@ -112,47 +117,221 @@ mod tests {
...
-        init_checkpoints(&mut genesis);
+        init_checkpoints(&mut genesis).unwrap();
+
+        let mut a: ProtocolState = Default::default();
+        let mut b: ProtocolState = Default::default();
+
+        gen_spot(&mut a);
+        gen_spot(&mut b);
+
         let mut c0: ProtocolStateChain = ProtocolStateChain(vec![]);
```
hanabi1224, line 249: let c0: ProtocolStateChain = ProtocolStateChain(vec![a]);

consensus/src/checkpoint.rs
```diff
@@ -112,47 +117,221 @@ mod tests {
...
+    #[test]
+    #[wasm_bindgen_test]
+    fn gen_spot_pair_short_misaligned_generates_pairs_of_states_in_short_fork_range() {
+        let mut genesis: ProtocolState = Default::default();
+        init_checkpoints(&mut genesis).unwrap();
+
+        let mut a: ProtocolState = Default::default();
+        let mut b: ProtocolState = Default::default();
+
+        gen_spot(&mut a);
```
hanabi1224, line 290: Why not making gen_spot a struct or trait method of ProtocolState

consensus/src/checkpoint.rs
```diff
@@ -112,47 +117,221 @@ mod tests {
...
+        gen_spot(&mut b);
+        a.body.consensus_state.epoch_count = Length(14);
+        b.body.consensus_state.epoch_count = Length(15);
+        let mut c0: ProtocolStateChain = ProtocolStateChain(vec![]);
```
hanabi1224, line 329: vec![a]

consensus/src/checkpoint.rs
```diff
@@ -74,8 +74,13 @@ pub fn is_short_range(
 #[cfg(test)]
 mod tests {
```
creativcoder, line 75: Bug: While going through the is_short_range function from the spec:

```rust
fn isShortRange(C1, C2) -> bool
{
    // Get consensus state from top blocks of each chain
    let S1 = cState(C1);
    let S2 = cState(C2);

    let check = | S1, S2 | {
      if S1.epoch_count == S2.epoch_count + 1  && epochSlot(S2) >= 2/3*slots_per_epoch {
        // S1 is one epoch ahead of S2 and S2 is not in the seed update range
        return S1.staking_epoch_data.lock_checkpoint == S2.next_epoch_data.lock_checkpoint
      }
      else {
        return false
      }
    };

    if S1.epoch_count == S2.epoch_count {
        // Simple case: blocks have same previous epoch, so compare previous epochs' lock_checkpoints
        return S1.staking_epoch_data.lock_checkpoint == S2.staking_epoch_data.lock_checkpoint
    }
    else {
        // Check for previous epoch case using both orientations
        return check(S1, S2) || check(S2, S1)
    }
}
```
it appears the impl here is missing the both way check:
`return check(S1, S2) || check(S2, S1)`

we should be checking it both ways

consensus/src/checkpoint.rs
```diff
@@ -112,47 +117,221 @@ mod tests {
...
+
+    fn gen_num_blocks_in_slots(slot_fill_rate: f64, slot_fill_rate_delta: f64, n: f64) -> i32 {
+        let min_blocks = n * f64::max(slot_fill_rate - slot_fill_rate_delta, 0.0);
+        let max_blocks = n * f64::min(slot_fill_rate + slot_fill_rate_delta, 1.0);
+        let num_blocks_in_slots = thread_rng().gen_range(min_blocks..max_blocks);
+        let num = convert(num_blocks_in_slots);
+        return num;
+    }
+
+    fn gen_spot(block: &mut ProtocolState) {
```
creativcoder, line 154: feels like this is missing steps as the one in ocaml: https://github.com/MinaProtocol/mina/blob/af76cb7980d5e81e704120290a850ea9c6f8522e/src/lib/consensus/proof_of_stake.ml#L4250 
?

creativcoder: As an added note: we want to be writing the test cases in our own way and not have to follow the code structure as in ocaml impl. Currently the test cases feels like we're porting the test cases.
What we can do is first arrive at a generic and code agnostic description of what the test cases are trying to achieve by reading the ocaml impl, and then re-write them here in our own way.

consensus/src/checkpoint.rs
```diff
@@ -112,47 +117,221 @@ mod tests {
...
+        // staking_epoch == root_epoch, next_staking_epoch == root_epoch + 1
+        // New state slot and length.
+        // New total currency for state.
+        // New epoch data for staking and next epochs.
+        // New block reward information (unused in chain selection).
+        let consensus_state = ConsensusState::new();
+        block.body.consensus_state = consensus_state;
+    }
+
+    fn gen_spot_pair_common_checkpoints(
```
creativcoder, line 166: this helper function eventually calls `gen_spot` in the ocaml impl at the end: https://github.com/MinaProtocol/mina/blob/af76cb7980d5e81e704120290a850ea9c6f8522e/src/lib/consensus/proof_of_stake.ml#L4250
which seems missing in our impl.

consensus/src/checkpoint.rs
polarbeargo: Our gen_spot is different in Mina Ocaml. Our gen_spot() is directly new a, b state then perform operations on the new a, b equal state with different conditions such as `gen_spot_pair_common_checkpoints` for `pair_short_aligned`, `pair_short_misaligned` and `pair_long`

consensus/src/checkpoint.rs
polarbeargo: Same as the above if we new a, b equal state first by calling `gen_spot` and performed operations on it already why do we need to call gen_spot again at the end of `gen_spot_pair_common_checkpoints`? These two functions overlap too much. This is my confusion.

consensus/src/checkpoint.rs
polarbeargo: I've implemented `return check(S1, S2) || check(S2, S1)` recursively but encountered an unreachable expression. And encounter stack overflows as the following. That's why I made this change:
Screenshot (271)

consensus/src/checkpoint.rs
polarbeargo: Then we don't need to use push here?

consensus/src/checkpoint.rs
polarbeargo: @creativcoder thank you for patiently waiting. Resolved.

base/src/consensus_state.rs
polarbeargo: Can't use in this case will raise conflict
Screenshot (387)

consensus/src/checkpoint.rs
polarbeargo: A const function cannot call a non-const function see [link](https://stackoverflow.com/questions/45249486/why-cant-const-functions-in-rust-make-calls-to-associated-functions)
Screenshot (389)

consensus/src/checkpoint.rs
polarbeargo: the slot is an integer type, I use slot as f64 just for gen_num_blocks_in_slots perform calculate with float number const.

consensus/src/checkpoint.rs
hanabi1224: You can do either
```rust
const a:f64 = 2_u64.pow(32) as f64;
```
or
```rust
const a:f64 = 4294967296_f64
```

base/src/consensus_state.rs
hanabi1224: Then u can remove the derived Default impl

polarbeargo: If remove derived Default will affect this method [line](https://github.com/ChainSafe/mina-rs/blob/6935fc291c3551498f12d1758f407bbef79068cb/base/src/consensus_state.rs#L75
Screenshot (394)
)

polarbeargo: If take this method out will cause test_protocol_state_chain_top to fail.
Screenshot (395)

consensus/src/checkpoint.rs
polarbeargo: Thank you for the excellent guidance 🥇

consensus/src/checkpoint.rs
```diff
@@ -57,15 +58,23 @@ pub fn is_short_range(
...
-
-    if s0.epoch_count.0 == s1.epoch_count.0 + 1 && c1.epoch_slot() >= Some(SLOTS_PER_EPOCH * 2 / 3)
-    {
-        Ok(s0_lock_checkpoint == s1_next_epoch_lock_checkpoint)
+        // Simple case: blocks have same previous epoch, so compare previous epochs' lock_checkpoints
+        Ok(s0_lock_checkpoint == s1_lock_checkpoint)
     } else {
-        Ok(false)
+        // Check for previous epoch case using both orientations
+        Ok(check(s0_lock_checkpoint, s1_next_epoch_lock_checkpoint)
```
creativcoder, line 77: Nice fix!

consensus/tests/checkpoint.rs
```diff
@@ -43,48 +50,201 @@ mod tests {
...
+
+        gen_spot(&mut a);
+        gen_spot(&mut b);
+        gen_spot_pair_common_checkpoints(&mut a, &mut b, 0);
+
+        let c0: ProtocolStateChain = ProtocolStateChain(vec![a]);
+        let c1: ProtocolStateChain = ProtocolStateChain(vec![b]);
+
         assert_eq!(is_short_range(&c0, &c1).unwrap(), true);
-        assert_eq!(is_short_range(&c1, &c0).unwrap(), true);
```
creativcoder, line 70: This assert seems commented

consensus/Cargo.toml
```diff
@@ -15,13 +15,14 @@ browser = []
 [dependencies]
 mina-crypto = {version = "0.1.0", path = "../crypto"}
 mina-rs-base = {version = "0.1.0", path = "../base"}
-
+snark-keys = {version = "0.1.0", path = "../snark-keys"}
```
creativcoder, line 18: Where is this being used?

consensus/tests/checkpoint.rs
polarbeargo: > Bug: While going through the is_short_range function from the spec:
> 
> ```rust
> fn isShortRange(C1, C2) -> bool
> {
>     // Get consensus state from top blocks of each chain
>     let S1 = cState(C1);
>     let S2 = cState(C2);
> 
>     let check = | S1, S2 | {
>       if S1.epoch_count == S2.epoch_count + 1  && epochSlot(S2) >= 2/3*slots_per_epoch {
>         // S1 is one epoch ahead of S2 and S2 is not in the seed update range
>         return S1.staking_epoch_data.lock_checkpoint == S2.next_epoch_data.lock_checkpoint
>       }
>       else {
>         return false
>       }
>     };
> 
>     if S1.epoch_count == S2.epoch_count {
>         // Simple case: blocks have same previous epoch, so compare previous epochs' lock_checkpoints
>         return S1.staking_epoch_data.lock_checkpoint == S2.staking_epoch_data.lock_checkpoint
>     }
>     else {
>         // Check for previous epoch case using both orientations
>         return check(S1, S2) || check(S2, S1)
>     }
> }
> ```
> 
> it appears the impl here is missing the both way check: `return check(S1, S2) || check(S2, S1)`
> 
> we should be checking it both ways


in the previous conversation, we resolved both ways check at is_short_range()  such that we don't need to check both ways again in the test.

base/src/protocol_state.rs
```diff
@@ -23,6 +23,38 @@ pub struct ProtocolConstants {
     pub genesis_state_timestamp: BlockTime,
 }
 
+pub struct ProtocolConstantsTemp {
```
creativcoder, line 26: It would be good to move this to consensus crate and name it something like `ConsensusConstants`: https://github.com/ChainSafe/mina-rs/pull/34/files#diff-1380b91b633a61c219d31250243405847967fde9e65ea2c4a4af6da87dc52753R13

consensus/Cargo.toml
polarbeargo: See [line](https://github.com/ChainSafe/mina-rs/blob/68acc0c4021d32de6f2f3af11cdc7927b2bfd975/consensus/tests/checkpoint.rs#L65) not sure we will need computes currency at height in the future tests like for staged ledger diff.
I'm reading [staged ledger diff](https://github.com/MinaProtocol/mina/blob/af76cb7980d5e81e704120290a850ea9c6f8522e/src/lib/staged_ledger_diff/staged_ledger_diff.ml) for #113

base/src/protocol_state.rs
polarbeargo: This temp struct was original from ProtocolConstans and is from [Constants](https://github.com/MinaProtocol/mina/blob/32f529d8d8d712a44ee75be66061ce08cbdc8924/docs/specs/consensus/README.md#3-constants). Should we move to consensus crate?

consensus/src/checkpoint.rs
polarbeargo: fixed

consensus/src/checkpoint.rs
polarbeargo: fixed

consensus/src/checkpoint.rs
polarbeargo: fixed

consensus/src/checkpoint.rs
```diff
@@ -112,47 +117,221 @@ mod tests {
...
+        init_checkpoints(&mut genesis).unwrap();
+
+        let mut a: ProtocolState = Default::default();
+        let mut b: ProtocolState = Default::default();
+
+        gen_spot(&mut a);
+        gen_spot(&mut b);
+
+        // Compute the root epoch position of `b`. This needs to be one epoch ahead of a
+        let default_slot_fill_rate = 0.65;
```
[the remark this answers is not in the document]
polarbeargo: fixed

consensus/src/checkpoint.rs
polarbeargo: fixed

consensus/Cargo.toml
polarbeargo: fixed

consensus/src/checkpoint.rs
polarbeargo: Currently, move gen_spot to common.rs such that can reuse gen_spot in any other test cases.

base/src/protocol_state.rs
polarbeargo: Fixed.

base/src/consensus_state.rs
hanabi1224: Sry but I still don't understand the root cause

consensus/src/checkpoint.rs
hanabi1224: Sry my point is not about reusability, but more on OO design

polarbeargo: Not sure should we put gen_spot as a trait for ProtocolState.

base/src/consensus_state.rs
polarbeargo: Affect this [line](https://github.com/ChainSafe/mina-rs/blob/68acc0c4021d32de6f2f3af11cdc7927b2bfd975/consensus/src/common.rs#L50)

consensus/src/checkpoint.rs
polarbeargo: `self` is a `&` reference, so the data it refers to cannot be written
Screenshot (400)
Screenshot (399)

polarbeargo: Fixed.

consensus/src/common.rs
```diff
@@ -26,6 +31,12 @@ where
     fn length(&self) -> usize;
     fn last_vrf(&self) -> Option<String>;
     fn state_hash(&self) -> Option<StateHash>;
+    fn gen_spot(block: &mut ProtocolState);
```
hanabi1224, line 34: Why have functions without self in this trait? It's not OO design and I don't really understand why they fit in the trait

base/src/consensus_state.rs
hanabi1224: Sry can u plz elaborate on why this is not possible? I think 'affect this line' is the result instead of the root cause,

consensus/src/common.rs
polarbeargo: previous review feedback mentioned why not integrate gen_spot in a trait or struct and you said it's more about OO design. In my previous merge, you speak about integrating these methods into Chain traits for better OO design. Then I just think maybe it's a good idea to put them into the Chain trait. In the case of is_short_range() there are two arguments how can they fit in self in Rust? It will be good if you can let me know what you prefer for a better OO design. A reference will be very nice.

polarbeargo: it's not good idea to use OO design here

base/src/consensus_state.rs
polarbeargo: The left blockchain_length: Length(0) compare right  blockchain_length: Length(1) if we use default implementation.
```
 left: `Some(ProtocolState { previous_state_hash: StateHash(BaseHash([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])), body: ProtocolStateBody { genesis_state_hash: StateHash(BaseHash([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])), blockchain_state: BlockchainState { staged_ledger_hash: StagedLedgerHash { non_snark: NonSnarkStagedLedgerHash { ledger_hash: LedgerHash(BaseHash([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])), aux_hash: AuxHash([]), pending_coinbase_aux: PendingCoinbaseAuxHash([]) }, pending_coinbase_hash: CoinBaseHash(BaseHash([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])) }, snarked_ledger_hash: SnarkedLedgerHash(BaseHash([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])), genesis_ledger_hash: SnarkedLedgerHash(BaseHash([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])), snarked_next_available_token: TokenId(0), timestamp: BlockTime(0) }, consensus_state: ConsensusState { blockchain_length: Length(0), epoch_count: Length(0), min_window_density: Length(77), sub_window_densities: [Length(1), Length(7), Length(7), Length(7), Length(7), Length(7), Length(7), Length(7), Length(7), Length(7), Length(7)], last_vrf_output: VrfOutputTruncated([13]), total_currency: Amount(805385692840039233), curr_global_slot: GlobalSlot { slot_number: GlobalSlotNumber(0), slots_per_epoch: Length(7140) }, global_slot_since_genesis: GlobalSlotNumber(0), staking_epoch_data: EpochData { ledger: EpochLedger { hash: LedgerHash(BaseHash([133, 246, 174, 16, 42, 48, 61, 19, 82, 234, 190, 184, 70, 31, 98, 164, 122, 104, 123, 109, 205, 231, 98, 51, 185, 32, 113, 104, 160, 172, 247, 44])), total_currency: Amount(805385692840039233) }, seed: EpochSeed(BaseHash([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])), start_checkpoint: StateHash(BaseHash([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])), lock_checkpoint: StateHash(BaseHash([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])), epoch_length: Length(1) }, next_epoch_data: EpochData { ledger: EpochLedger { hash: LedgerHash(BaseHash([133, 246, 174, 16, 42, 48, 61, 19, 82, 234, 190, 184, 70, 31, 98, 164, 122, 104, 123, 109, 205, 231, 98, 51, 185, 32, 113, 104, 160, 172, 247, 44])), total_currency: Amount(805385692840039233) }, seed: EpochSeed(BaseHash([37, 127, 117, 208, 167, 148, 99, 11, 218, 242, 217, 7, 37, 244, 201, 93, 161, 26, 189, 134, 47, 184, 162, 248, 138, 185, 76, 65, 250, 11, 141, 53])), start_checkpoint: StateHash(BaseHash([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])), lock_checkpoint: StateHash(BaseHash([232, 149, 55, 214, 162, 25, 166, 75, 189, 7, 127, 231, 219, 235, 171, 161, 183, 9, 227, 123, 211, 166, 117, 103, 24, 130, 127, 33, 8, 154, 124, 2])), epoch_length: Length(2) }, has_ancestor_in_same_checkpoint_window: true, block_stake_winner: PublicKey { poly: CompressedCurvePoint { x: [15, 72, 198, 91, 210, 95, 133, 243, 228, 234, 78, 254, 190, 183, 91, 121, 123, 215, 67, 96, 59, 224, 75, 78, 173, 132, 86, 152, 183, 107, 211, 49], is_odd: false } }, block_creator: PublicKey { poly: CompressedCurvePoint { x: [15, 72, 198, 91, 210, 95, 133, 243, 228, 234, 78, 254, 190, 183, 91, 121, 123, 215, 67, 96, 59, 224, 75, 78, 173, 132, 86, 152, 183, 107, 211, 49], is_odd: false } }, coinbase_receiver: PublicKey { poly: CompressedCurvePoint { x: [15, 72, 198, 91, 210, 95, 133, 243, 228, 234, 78, 254, 190, 183, 91, 121, 123, 215, 67, 96, 59, 224, 75, 78, 173, 132, 86, 152, 183, 107, 211, 49], is_odd: false } }, supercharge_coinbase: true }, constants: ProtocolConstants { k: Length(0), slots_per_epoch: Length(0), slots_per_sub_window: Length(0), delta: Delta(0), genesis_state_timestamp: BlockTime(0) } } })`,
 right: `Some(ProtocolState { previous_state_hash: StateHash(BaseHash([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])), body: ProtocolStateBody { genesis_state_hash: StateHash(BaseHash([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])), blockchain_state: BlockchainState { staged_ledger_hash: StagedLedgerHash { non_snark: NonSnarkStagedLedgerHash { ledger_hash: LedgerHash(BaseHash([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])), aux_hash: AuxHash([]), pending_coinbase_aux: PendingCoinbaseAuxHash([]) }, pending_coinbase_hash: CoinBaseHash(BaseHash([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])) }, snarked_ledger_hash: SnarkedLedgerHash(BaseHash([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])), genesis_ledger_hash: SnarkedLedgerHash(BaseHash([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])), snarked_next_available_token: TokenId(0), timestamp: BlockTime(0) }, consensus_state: ConsensusState { blockchain_length: Length(1), epoch_count: Length(0), min_window_density: Length(77), sub_window_densities: [Length(1), Length(7), Length(7), Length(7), Length(7), Length(7), Length(7), Length(7), Length(7), Length(7), Length(7)], last_vrf_output: VrfOutputTruncated([13]), total_currency: Amount(805385692840039233), curr_global_slot: GlobalSlot { slot_number: GlobalSlotNumber(0), slots_per_epoch: Length(7140) }, global_slot_since_genesis: GlobalSlotNumber(0), staking_epoch_data: EpochData { ledger: EpochLedger { hash: LedgerHash(BaseHash([133, 246, 174, 16, 42, 48, 61, 19, 82, 234, 190, 184, 70, 31, 98, 164, 122, 104, 123, 109, 205, 231, 98, 51, 185, 32, 113, 104, 160, 172, 247, 44])), total_currency: Amount(805385692840039233) }, seed: EpochSeed(BaseHash([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])), start_checkpoint: StateHash(BaseHash([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])), lock_checkpoint: StateHash(BaseHash([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])), epoch_length: Length(1) }, next_epoch_data: EpochData { ledger: EpochLedger { hash: LedgerHash(BaseHash([133, 246, 174, 16, 42, 48, 61, 19, 82, 234, 190, 184, 70, 31, 98, 164, 122, 104, 123, 109, 205, 231, 98, 51, 185, 32, 113, 104, 160, 172, 247, 44])), total_currency: Amount(805385692840039233) }, seed: EpochSeed(BaseHash([37, 127, 117, 208, 167, 148, 99, 11, 218, 242, 217, 7, 37, 244, 201, 93, 161, 26, 189, 134, 47, 184, 162, 248, 138, 185, 76, 65, 250, 11, 141, 53])), start_checkpoint: StateHash(BaseHash([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])), lock_checkpoint: StateHash(BaseHash([232, 149, 55, 214, 162, 25, 166, 75, 189, 7, 127, 231, 219, 235, 171, 161, 183, 9, 227, 123, 211, 166, 117, 103, 24, 130, 127, 33, 8, 154, 124, 2])), epoch_length: Length(2) }, has_ancestor_in_same_checkpoint_window: true, block_stake_winner: PublicKey { poly: CompressedCurvePoint { x: [15, 72, 198, 91, 210, 95, 133, 243, 228, 234, 78, 254, 190, 183, 91, 121, 123, 215, 67, 96, 59, 224, 75, 78, 173, 132, 86, 152, 183, 107, 211, 49], is_odd: false } }, block_creator: PublicKey { poly: CompressedCurvePoint { x: [15, 72, 198, 91, 210, 95, 133, 243, 228, 234, 78, 254, 190, 183, 91, 121, 123, 215, 67, 96, 59, 224, 75, 78, 173, 132, 86, 152, 183, 107, 211, 49], is_odd: false } }, coinbase_receiver: PublicKey { poly: CompressedCurvePoint { x: [15, 72, 198, 91, 210, 95, 133, 243, 228, 234, 78, 254, 190, 183, 91, 121, 123, 215, 67, 96, 59, 224, 75, 78, 173, 132, 86, 152, 183, 107, 211, 49], is_odd: false } }, supercharge_coinbase: true }, constants: ProtocolConstants { k: Length(0), slots_per_epoch: Length(0), slots_per_sub_window: Length(0), delta: Delta(0), genesis_state_timestamp: BlockTime(0) } } })`
```

polarbeargo: At [line](https://github.com/ChainSafe/mina-rs/blob/6935fc291c3551498f12d1758f407bbef79068cb/consensus/tests/common.rs#L50)
The one below line 50 passes the test
Screenshot (402)

polarbeargo: Fixed.

consensus/src/checkpoint.rs
polarbeargo: Fixed

consensus/Cargo.toml
creativcoder: Feels like we should tackle that as a separate PR

consensus/src/common.rs
```diff
@@ -79,3 +79,14 @@ impl Chain<ProtocolState> for ProtocolStateChain {
         self.top().map(|s| s.hash())
     }
 }
+
+pub fn gen_spot(block: &mut ProtocolState) {
```
creativcoder, line 83: If this function just does one line, better to inline it where it's called and maybe add a comment hinting that this is part of gen_spot

polarbeargo: Fixed.

consensus/tests/checkpoint.rs
```diff
@@ -5,8 +5,14 @@
 mod tests {
     use mina_consensus::{checkpoint::*, common::*};
     use mina_crypto::{hash::*, prelude::*};
+    use mina_rs_base::numbers::Length;
     use mina_rs_base::types::*;
+    use rand::{thread_rng, Rng};
```
willemolding, line 10: In my opinion tests should be deterministic and introducing randomness can break this.

If we absolutely have to use randomness we should make sure it is seeded in the code (which `ThreadRng` is not). Something like:
```rust
use rand::{SeedableRng, rngs::SmallRng};

// snip

SmallRng::seed_from_u64(SEED);
```
where `SEED` is hard-coded somewhere.

Much better would be to use hard-coded values and eliminate this dependency altogether

snark-keys/src/header.rs
```diff
@@ -45,12 +45,12 @@ pub enum KeyIdentifier {
     TransactionSnark,
 }
 
-#[derive(Debug, Deserialize)]
+#[derive(Debug, Default, Deserialize)]
```
willemolding, line 48: Hopefully we can remove these changes if we can drop the dependency

willemolding (changes requested):

base/src/consensus_state.rs
```diff
@@ -71,8 +76,127 @@ pub struct ConsensusState {
     pub supercharge_coinbase: bool,
 }
 
+pub struct ConsensusConstants {
```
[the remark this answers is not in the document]
polarbeargo: Yes, @creativcoder said we need independent these variants out of protocol constant so we move these variants to ConsensusConstants. If remove them clippy will throw warnings.

consensus/tests/checkpoint.rs
polarbeargo: It's from [here ](https://github.com/MinaProtocol/mina/blob/af76cb7980d5e81e704120290a850ea9c6f8522e/src/lib/consensus/proof_of_stake.ml#L3960) where use [ Core.Int.gen_incl](http://212.47.237.247/packages/core/v0_9_2/Core/Int/index.html)

consensus/tests/checkpoint.rs
```diff
@@ -5,8 +5,14 @@
...
+    use snark_keys::ConstraintConstants;
```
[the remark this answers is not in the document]
polarbeargo: Sure. I can remove this [function](https://github.com/ChainSafe/mina-rs/blob/b28332036b8f1165141c6907053dd8a60316c36b/consensus/tests/checkpoint.rs#L63) to remove the dependency of the snark_keys crate.

base/src/consensus_state.rs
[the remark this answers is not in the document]
willemolding: No worries maybe try and fix this later on

consensus/tests/checkpoint.rs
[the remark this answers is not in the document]
willemolding: I'd do it if it doesn't decrease test coverage

snark-keys/src/header.rs
polarbeargo: Fixed

consensus/tests/checkpoint.rs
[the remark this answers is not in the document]
polarbeargo: Fixed

consensus/tests/checkpoint.rs
polarbeargo: Won't it be weird a test is always around mid (max_blocks + min_blocks) / 2 point?
Sure the test will pass but every time you are run the same hard code value.

polarbeargo: Won't it be weird to use the same hard code values every time while running tests?

polarbeargo: The original tests in OCaml code were using a property-based testing library to generate these values, but we cannot use that here because of wasm incompatibility.

consensus/src/checkpoint.rs
```diff
@@ -48,6 +48,7 @@ pub fn is_short_range(
     c0: &ProtocolStateChain,
     c1: &ProtocolStateChain,
 ) -> Result<bool, ConsensusErrTyp> {
+    // Get consensus state from top blocks of each chain
```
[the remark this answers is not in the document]
polarbeargo: Fixed.

consensus/src/checkpoint.rs
```diff
@@ -57,15 +58,23 @@ pub fn is_short_range(
     let s0_lock_checkpoint = &s0.staking_epoch_data.lock_checkpoint;
     let s1_lock_checkpoint = &s1.staking_epoch_data.lock_checkpoint;
     let s1_next_epoch_lock_checkpoint = &s1.next_epoch_data.lock_checkpoint;
+    let check = |s0_lock_checkpoint, s1_next_epoch_lock_checkpoint| {
```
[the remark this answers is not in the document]
polarbeargo: It's from [spec](https://github.com/MinaProtocol/mina/blob/32f529d8d8d712a44ee75be66061ce08cbdc8924/docs/specs/consensus/README.md#532-short-range-fork-check)

consensus/tests/checkpoint.rs
polarbeargo: Fixed. Not use rand at all.

base/src/consensus_state.rs
```diff
@@ -81,9 +86,129 @@ pub struct ConsensusState {
...
+    pub genesis_state_timestamp: BlockTime,
+    /// Sub windows within a window
+    pub sub_windows_per_window: Length,
+    /// Number of slots before minimum density is used in chain selection
+    pub grace_period_end: Length,
+}
+
+impl Default for ConsensusConstants {
+    fn default() -> Self {
+        // TODO: read from config
```
noot, line 109: let's open an issue for this if there isn't one

base/src/consensus_state.rs
```diff
@@ -81,9 +86,129 @@ pub struct ConsensusState {
...
+        Self {
+            k: Length(290),
+            slots_per_epoch: 7140.into(),
+            slots_per_sub_window: 7.into(),
+            delta: 0.into(),
+            genesis_state_timestamp: BlockTime(1615939200000),
+            sub_windows_per_window: 11.into(),
+            grace_period_end: Length(1440),
+        }
+    }
```
noot, line 119: where are these values from?

base/src/consensus_state.rs
```diff
@@ -81,9 +86,129 @@ pub struct ConsensusState {
...
+impl Default for ConsensusState {
+    // TODO: read from config
+    fn default() -> Self {
+        let total_currency = Amount(805385692840039233);
+        let staking_epoch_data = {
+            let mut data = EpochData::default();
+
+            data.epoch_length.0 = 1;
+            data.ledger.hash =
+                LedgerHash::from_base58("jx7buQVWFLsXTtzRgSxbYcT8EYLS8KCZbLrfDcJxMtyy4thw2Ee")
```
noot, line 138: same for this function, let's document where the values are from

consensus/src/checkpoint.rs
```diff
@@ -44,10 +44,14 @@ pub fn init_checkpoints(genesis: &mut ProtocolState) -> Result<(), ConsensusErrT
     Ok(())
 }
 
+/// A fork is considered short-range if either
+/// the fork point of the candidate chains are in the same epoch
+/// or the fork point is in the previous epoch with the same lock_checkpoint
```
noot, line 49: can the fork point be anywhere in the previous epoch? i thought it was only after the lock_checkpoint

consensus/src/checkpoint.rs
```diff
@@ -57,15 +61,23 @@ pub fn is_short_range(
     let s0_lock_checkpoint = &s0.staking_epoch_data.lock_checkpoint;
     let s1_lock_checkpoint = &s1.staking_epoch_data.lock_checkpoint;
     let s1_next_epoch_lock_checkpoint = &s1.next_epoch_data.lock_checkpoint;
+    let check = |s0_lock_checkpoint, s1_next_epoch_lock_checkpoint| {
+        if s0.epoch_count.0 == s1.epoch_count.0 + 1
+            && c1.epoch_slot() >= Some(SLOTS_PER_EPOCH * 2 / 3)
+        {
+            // S1 is one epoch ahead of S2 and S2 is not in the seed update range
```
noot, line 68: should this comment say s0 and s1?

consensus/tests/checkpoint.rs
```diff
@@ -5,8 +5,12 @@
 mod tests {
     use mina_consensus::{checkpoint::*, common::*};
     use mina_crypto::{hash::*, prelude::*};
+    use mina_rs_base::numbers::Length;
     use mina_rs_base::types::*;
     use wasm_bindgen_test::*;
+    const DEFAULT_SLOT_FILL_RATE: f64 = 0.65;
+    const DEFAULT_SLOT_FILL_RATE_DELTA: f64 = 0.15;
+    const TRANS: f64 = 4294967296_f64;
```
noot, line 13: where are these from?

noot: also a comment for each describing what they are would be good!

consensus/tests/checkpoint.rs
```diff
@@ -43,48 +47,194 @@ mod tests {
...
+        // Constraining the second state to have a greater blockchain length than the
+        // first, we need to constrain the first blockchain length such that there is some room
+        // leftover in the epoch for at least 1 more block to be generated.
+        let _blockchain_length_relativity = Some(&a);
+        let max_epoch_slot = match _blockchain_length_relativity {
+            Some(_blockchain_length_relativity) => {
+                a.body.consensus_state.curr_global_slot.slots_per_epoch.0 - 4
+            } // -1 to bring into inclusive range, -3 to provide 2 slots of fudge room
+            None => a.body.consensus_state.curr_global_slot.slots_per_epoch.0 - 1, // -1 to bring into inclusive range
+        };
```
noot, line 106: when would the `None` case ever be hit here? since `_blockchain_length_relativity = Some(&a)` it looks like it's always `Some`

consensus/tests/checkpoint.rs
```diff
@@ -43,48 +47,194 @@ mod tests {
...
+        let mut b: ProtocolState = Default::default();
+
+        a.body.consensus_state = ConsensusState::default();
+        b.body.consensus_state = ConsensusState::default();
+
+        a.body.consensus_state.epoch_count = Length(14);
+        b.body.consensus_state.epoch_count = Length(15);
+        let c0: ProtocolStateChain = ProtocolStateChain(vec![a]);
+        let c1: ProtocolStateChain = ProtocolStateChain(vec![b]);
+
```
noot, line 237: shouldn't the slot numbers (`epoch_slot`) be set here as well? won't they have the same slot number here?

willemolding (approved):
So happy now the randomness is gone :P

base/src/consensus_state.rs
polarbeargo: Use in this [line](https://github.com/ChainSafe/mina-rs/blob/70098ebe162625ac8be18fec3c5da589fedd417b/consensus/tests/checkpoint.rs#L211) and this [line](https://github.com/ChainSafe/mina-rs/blob/70098ebe162625ac8be18fec3c5da589fedd417b/consensus/tests/checkpoint.rs#L51) is origin from ProrocolConstant. See [Discussion](https://github.com/ChainSafe/mina-rs/pull/93#discussion_r780879450)

base/src/consensus_state.rs
polarbeargo: From our previous codebase for the new default value. Will document these

consensus/src/checkpoint.rs
polarbeargo: See [spec](https://github.com/MinaProtocol/mina/blob/32f529d8d8d712a44ee75be66061ce08cbdc8924/docs/specs/consensus/README.md#53-decentralized-checkpointing)
Screenshot (419)

consensus/src/checkpoint.rs
polarbeargo: Fixed

consensus/tests/checkpoint.rs
polarbeargo: See the [Mina Ocaml](https://github.com/MinaProtocol/mina/blob/af76cb7980d5e81e704120290a850ea9c6f8522e/src/lib/consensus/proof_of_stake.ml#L3948) no any document and description. Will ask O1 lab. And TRANS is define constant for 2f64.powi(32) for this [func](https://github.com/ChainSafe/mina-rs/blob/38edaa280d2c25f714481b908b9fb01209b67d2b/consensus/tests/checkpoint.rs#L72)

consensus/tests/checkpoint.rs
polarbeargo: We use new a, b state then directly use it as _blockchain_length_relativity  and there's never [blockchain_length_relativity](https://github.com/MinaProtocol/mina/blob/af76cb7980d5e81e704120290a850ea9c6f8522e/src/lib/consensus/proof_of_stake.ml#L4258) input from origin mina OCaml just a, b state see [url](https://github.com/MinaProtocol/mina/blob/af76cb7980d5e81e704120290a850ea9c6f8522e/src/lib/consensus/proof_of_stake.ml#L4414)

consensus/tests/checkpoint.rs
polarbeargo: No, we already modify into this [line ](https://github.com/ChainSafe/mina-rs/blob/38edaa280d2c25f714481b908b9fb01209b67d2b/base/src/consensus_state.rs#L191) not use the previous way.

base/src/consensus_state.rs
polarbeargo: Opend [#127 ](https://github.com/ChainSafe/mina-rs/issues/127)

consensus/tests/checkpoint.rs
polarbeargo: Fixed.

consensus/tests/checkpoint.rs
creativcoder: `Ascending and `Equal are just enum variants they define in the code. See https://ocaml.org/learn/tutorials/labels.html "More variants" section at the bottom

creativcoder: so the value of blockchain_length_relativity is not `&mut ProtocolState`, but rather either `Ascending or `Equal based on the ocaml code.

polarbeargo: I have no idea how they decide what is Ascending or Equal in Mina Ocaml. And there's no blockchain_relativity input from Mina Ocaml. Should it means default will be None?

base/src/consensus_state.rs
polarbeargo: Fixed document [Initialize consensus JSON template](https://github.com/MinaProtocol/mina/blob/32f529d8d8d712a44ee75be66061ce08cbdc8924/docs/specs/consensus/README.md#61-initialize-consensus)

base/src/consensus_state.rs
polarbeargo: Fixed document [Initialize consensus JSON template](https://github.com/MinaProtocol/mina/blob/32f529d8d8d712a44ee75be66061ce08cbdc8924/docs/specs/consensus/README.md#61-initialize-consensus)

consensus/tests/checkpoint.rs
creativcoder: > The blockchain_relativety is a state please see the above links I attached

I think you mean `blockchain_length_relativity` and not `blockchain_relativity` as there's no `blockchain_relativity` in code.
ok, let me help see it better. If you take a look at https://github.com/MinaProtocol/mina/blob/af76cb7980d5e81e704120290a850ea9c6f8522e/src/lib/consensus/proof_of_stake.ml#L4140 clearly `blockchain_length_relativity` is being `match`ed with the `Ascending` and `Equal` variant they have defined. If `blockchain_length_relativity` were a `ProtocolState`, then the match arm would have matched with either an instance of protocol state or some of its fields, which isn't the case. hope that helps :)
 

>  I have no idea how they decide what is Ascending or Equal

Those are just enum variants they define, as i replied above. Asceding or Equal is totally different than > or = and they are not at all related here. They are a new custom enum type they have defined.

polarbeargo: I already repeat the same thing. Hope you don't repeat what I'm saying again, thank you.

polarbeargo: We use default a, b state then directly use a state as _blockchain_length_relativity root base to create relativity for max_epoch_slot because there's no blockchain_length_relativity in the input argument just a, b state in the second screenshot.
Screenshot (424)
Screenshot (425)
.

polarbeargo: They only set `blockchain_length_relativity` in the selection case which is not include in the is_short_range test case.

polarbeargo: It should always be none in this is_short_range test cases PR. No need to handle `_blockchain_length_relativity`

consensus/src/checkpoint.rs
polarbeargo: From the spec, I only know the 2nd condition matches when the fork point is in the previous epoch with the same lock_checkpoint.

creativcoder:
Closing this and starting afresh here: https://github.com/ChainSafe/mina-rs/pull/151

PR #103, opened 2021-12-15
Put binprot loosely-typed deserialization behind optional feature

hanabi1224:
**Summary of changes**
- To address https://github.com/ChainSafe/mina-rs/pull/94#issuecomment-988041836
- Add feature `layout` to bin-prot crate
- Turn off `layout` feature for non-testing crates
- Test block deserialization (w/o layout) with random data to make sure it fails gracefully

@jspada plz review

Closes 

### DIFF
--- a/protocol/test-serialization/src/fuzz.rs
+++ b/protocol/test-serialization/src/fuzz.rs
@@ -1,0 +1,61 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0
+
+#[cfg(test)]
+mod tests {
+    use crate::fuzz_test;
+    use bin_prot::Deserializer;
+    use mina_rs_base::types::*;
+    use rand::prelude::*;
+    use serde::Deserialize;
+    use wasm_bindgen_test::*;
+
+    #[test]
+    #[wasm_bindgen_test]
+    fn test_corrupted_deserialization() {
+        fuzz_test!(
+            ExternalTransition
+
+            ProtocolState
+            ProtocolStateBody
+            BlockchainState
+            ConsensusState
+            ProtocolConstants
+
+            ProtocolStateProof
+            ProofStatement
+            PrevEvals
+            Proof
+            ProofMessages
+            ProofOpenings
+
+            StagedLedgerDiff
+            StagedLedgerDiffTuple
+            StagedLedgerPreDiffTwo
+
+            DeltaTransitionChainProof
+
+            Option<ProtocolVersion>
+            ()
+        );
+    }
+
+    #[macro_export]
+    macro_rules! fuzz_test {
+        ($($ty: ty) *) => {
+            $(
+                let mut rng = rand::thread_rng();
+                for _i in 0..5 {
+                    let mut bytes = vec![0; rng.gen_range((1024 * 1024)..(50 * 1024 * 1024))];
+                    rng.try_fill_bytes(&mut bytes).unwrap();
+                    let et: anyhow::Result<$ty> = (|| {
+                        let mut de = Deserializer::from_reader(bytes.as_slice());
+                        let et: $ty = Deserialize::deserialize(&mut de)?;
+                        Ok(et)
+                    })();
+                    et.unwrap_err();
+                }
+            )*
+        };
+    }
+}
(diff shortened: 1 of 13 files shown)


protocol/test-serialization/src/lib.rs
```diff
@@ -500,6 +501,22 @@ mod tests {
         test_roundtrip(&block.value, block.bytes.as_slice());
     }
 
+    #[test]
+    #[wasm_bindgen_test]
+    fn test_corrupted_deserialization() {
+        let mut rng = rand::thread_rng();
```
polarbeargo, line 507: Awsome work using rand generates random data for more robust test 👍 !

jspada: Looks great

jspada: I think you probably need to create individual unit tests for each sub-type of `ExternalTransition` to assure each of those individually fails on corrupt data.

Strictly speaking, this test does not prove that *some* corruption to these members won't go undetected.  As an extreme example, it could be the case that this top-level `ExternalTransition` negative test only passes because corruption of a single members is being detected (while the rest are not).   That could mean that another test case with corruption *only in those other members* could still fail.  Thus, I think you ought to create a corruption test for each sub-type.

hanabi1224: Added coverage for more compound structs. Please note that some simple structs can be deserialized into with no error as some of the checks will happen later when the bin-prot Serializable struct is converted into another equivalent struct for calculations, e.g. it does not ensure random public keys can be converted into BigInteger256 successfully at this stage, the check will happen upon the conversion, which is pending mina crypto refactoring

protocol/test-serialization/src/fuzz.rs
```diff
@@ -0,0 +1,61 @@
...
+            Option<ProtocolVersion>
+            ()
+        );
+    }
+
+    #[macro_export]
+    macro_rules! fuzz_test {
+        ($($ty: ty) *) => {
+            $(
+                let mut rng = rand::thread_rng();
```
jspada, line 47: Each time this runs it will produce different random bytes.  Therefore, if it fails and you try to fix and rerun it, then the you will not get the same test case.  It's usually better to be deterministic, i.e., seed the Rng so that it produces the same random bytes each time it's run.   For example like this `let mut rng = StdRng::from_seed([0; 32]);`

protocol/test-serialization/src/fuzz.rs
```diff
@@ -0,0 +1,61 @@
...
+                for _i in 0..5 {
+                    let mut bytes = vec![0; rng.gen_range((1024 * 1024)..(50 * 1024 * 1024))];
```
jspada, line 49: This `gen_range` is producing a large amount of bytes, bigger than any type.

This is a good test, you should probably also have some tests where the size of the random bytes is capped at the size of the corresponding object you are deserialising?  This way we know the test is not just passing because of a size check.

willemolding (approved):
Nice one! 
One suggestion I have is to rename the feature from `layout` to `loose_deserialization`.

Also add documentation about the features to the README.md

polarbeargo (approved):
Awsome work 🚀 !

protocol/test-serialization/src/fuzz.rs
hanabi1224: Fixed

protocol/test-serialization/src/fuzz.rs
hanabi1224: The binary is deserialized as a stream and in the current design, it does not require the stream to reach its end to allow non-exhaustive struct definition(although the omitted field has to be the tailing ones), I've added another `ensure_non_exhaustive_deserialization` test to demonstrate that.

protocol/test-serialization/src/fuzz.rs
jspada: Nice - thanks

protocol/test-serialization/src/fuzz.rs
jspada: Thanks 👍🏼

protocol/test-serialization/src/lib.rs
jspada: 👍🏼

PR #123, opened 2022-01-13
[M4] mina merkle tree

### LINKED ISSUE #129: M4 - implement merkle tree structure
**Task summary**

- mina uses a standard merkle tree structure for state storage
- implement standard merkle tree structure with the specified hashing algorithm, hashing algorithm used is poseidon, see https://github.com/MinaProtocol/c-reference-signer - there is probably an existing rust implementation for this
- we do **not** need to implement account structure at this point, we can leave the tree contents generic

- hasher is initialized with the merkle tree height: https://github.com/MinaProtocol/mina/blob/develop/src/lib/mina_base/ledger_hash.ml#L37
- also initialized with a salt prefix: https://github.com/MinaProtocol/mina/blob/develop/src/lib/hash_prefix_states/hash_prefix_states.ml#L16
- empty parts of the tree are empty hashes https://github.com/MinaProtocol/mina/blob/develop/src/lib/mina_base/ledger_hash.ml#L43 (`Field.zero`)

potentially closed by #123

hanabi1224:
**Summary of changes**
- https://github.com/o1-labs/snarky/blob/master/src/base/merkle_tree.ml

Closes  #129 


merkle/src/metadata.rs
```diff
@@ -0,0 +1,43 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0
+
+/// Metadata of a give tree node, including index and depth
+/// in the merkle tree it belongs to, which can be used for
+/// calculating hash
+#[derive(Debug, Clone, Default)]
+pub struct MerkleTreeNodeMetadata {
+    index: usize,
```
creativcoder, line 9: is this missing a depth field as described by the doc?

creativcoder: ok i see it's being derived from index. nvm

merkle/src/tree_impl.rs
```diff
@@ -0,0 +1,182 @@
...
+        (size as f64).log2().ceil() as u32
+    }
+}
+
+fn calculate_node_count(depth: u32) -> usize {
+    2_usize.pow(depth) - 1
+}
+
+fn calculate_parent_index(index: usize) -> usize {
+    (index - 1) / 2
```
creativcoder, line 164: `index.saturating_sub(1)` as `index-1` will panic, if `index` is `0`

merkle/tests/tree.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    struct TestMerger {}
+
+    impl MerkleMerger for TestMerger {
+        type Hash = i64;
+        fn merge(
+            left: &Option<(Self::Hash, MerkleTreeNodeMetadata)>,
+            right: &Option<(Self::Hash, MerkleTreeNodeMetadata)>,
+        ) -> Option<Self::Hash> {
+            if let Some((left, _)) = left {
```
creativcoder, line 44: this could be written like this:
```rust
left.as_ref().and_then(|(left, _)| right.as_ref().and_then(|(right, _)| Some(left + right)))
```

but not sure if that's more readable.

merkle/src/tree_impl.rs
```diff
@@ -0,0 +1,182 @@
...
+        for leaf in items
+            .into_iter()
+            .enumerate()
+            .map(|(i, item)| (THasher::hash(&item, MerkleTreeNodeMetadata::new(i)), item))
+        {
+            self.leafs.push(leaf);
+        }
+    }
+
+    fn add(&mut self, item: Self::Item) {
```
creativcoder, line 129: is add supposed to create a new `vec![]` everytime for adding the passed `item`

merkle/src/lib.rs
```diff
@@ -0,0 +1,23 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0
+
+#![deny(warnings)]
+#![deny(missing_docs)]
+
+//! mina-merkle crate provides traits and data structure implementations for
+//! in-memory, persistent, maskable and masking mina merkle tree
```
creativcoder, line 8: A little dive into how all of the maskable and merkle works in the context of mina would be enlightening once we have the details.

hanabi1224: maskable / masking features are trait only, will add more context in docs when I start implementing them, not in this PR tho.

merkle/src/tree_impl.rs
hanabi1224: No it creates a new `vec![]` only when tree depth / height is changed, in that case all hashes need to be recalculated anyway

merkle/src/tree_impl.rs
hanabi1224: This is a private function and the index is not supposed to be 0. If that ever happens I guess it's better to panic than returning unexpected values.

hanabi1224: Let me add 'debug_assert!(index > 0);' here

merkle/tests/tree.rs
hanabi1224: Tried this but tests failed, looks like this does not handle line 48 tho,

creativcoder: oh that can be handled with chaining `or` after right's and_then:
```rust
left.as_ref().and_then(|(left, _)| right.as_ref().and_then(|(right, _)| Some(left + right)).or(Some(*left)))
```

but i guess it's okay to keep it the usual if else way.

merkle/src/tree_impl.rs
```diff
@@ -0,0 +1,183 @@
...
+        let protential_depth = calculate_depth(capacity);
+        let protential_node_count = calculate_node_count(protential_depth);
+        Self {
+            leafs: Vec::with_capacity(capacity),
+            nodes: Vec::with_capacity(protential_node_count),
+            ..Default::default()
+        }
+    }
+
+    fn clear_dirty_hashes(&mut self, leaf_index: usize) {
```
noot, line 46: what does this do? a comment would be good!

merkle/src/tree_impl.rs
```diff
@@ -0,0 +1,183 @@
...
+            parent = calculate_parent_index(parent);
+            if self.nodes[parent].is_some() {
+                self.nodes[parent] = None;
+            } else {
+                break;
+            }
+        }
+    }
+
+    fn calculate_hash_if_needed(
```
noot, line 58: same here! I'm also curious as to what happens if there's an odd number of leaf nodes?

merkle/tests/tree.rs
```diff
@@ -0,0 +1,137 @@
...
+    fn mina_merkle_tree_tests_129() {
+        test_mina_merkle_tree(129, 8256, 8);
+    }
+
+    #[test]
+    fn mina_merkle_tree_tests_188() {
+        test_mina_merkle_tree(188, 17578, 8);
+    }
+
+    fn test_mina_merkle_tree(n: usize, expected_root_hash: i64, expected_depth: u32) {
```
noot, line 97: why is the root hash an i64? what hashing algorithm does mina use?

merkle/src/tree_impl.rs
hanabi1224: Fixed

merkle/src/tree_impl.rs
hanabi1224: Fixed. In the ocaml merkle tree implementation, it does not handle the even-odd case, I think it just delegates that logic to the merger.

merkle/tests/tree.rs
hanabi1224: This is only for testing, in the context of https://github.com/o1-labs/snarky/blob/master/src/base/merkle_tree.ml hash and merge algorithms are not specified, mina specific hasher merger implementations will come in subsequent PR(s)

merkle/src/metadata.rs
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    /// 0-based Index of the tree node
+    pub fn index(&self) -> usize {
+        self.index
+    }
+
+    /// 0-based depth of the tree node
+    pub fn depth(&self) -> u32 {
+        (self.index as f64 + 2.).log2().ceil() as u32 - 1
```
willemolding, line 26: This assumes binary Merkle tree right? Please see my totally badass use of const generics for a tree of any degree
```rust
#[derive(Debug, Clone, Default)]
pub struct MerkleTreeNodeMetadata<const DEGREE: usize> {
    index: usize,
}

impl<const DEGREE: usize> MerkleTreeNodeMetadata<DEGREE> {
    /// Creates a new instance of [MerkleTreeNodeMetadata] with
    /// its index in the merkle tree
    pub fn new(index: usize) -> Self {
        Self { index }
    }

    /// 0-based Index of the tree node
    pub fn index(&self) -> usize {
        self.index
    }

    /// 0-based depth of the tree node
    pub fn depth(&self) -> usize {
        ((self.index + DEGREE) as f64).log(ARITY as f64).ceil() as usize - 1
    }
}
```

DISCLAIMER: I have tested this builds but not that it works :P

merkle/src/tree.rs
```diff
@@ -0,0 +1,22 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0
+
+/// Trait for implementing general binary merkle tree
+pub trait MerkleTree {
```
willemolding, line 5: Could also add the const generics here

merkle/src/metadata.rs
willemolding: This is my favorite new Rust feature

creativcoder (approved):

crypto/src/hash/prefixes.rs
```diff
@@ -93,7 +93,28 @@ mod tests {
     }
 
     #[test]
-    fn make_merkle_tree_hash() {
+    fn make_merkle_tree_hash_3() {
```
jspada, line 96: In `create()` above, please discontinue use of single letter variable names.

Optionally, since the prefix is always a string, another way could be

```rust
fn create(prefix: &str) -> [u8; PREFIX_BYTE_LEN] {
    let prefix = &prefix[..std::cmp::min(prefix.len(), PREFIX_BYTE_LEN)];
    format!("{:*<PREFIX_BYTE_LEN$}", prefix)
        .as_bytes()
        .try_into()
        .expect("bad length")
}
```
N.b. bad length is not possible.

jspada: Also, could really use some rustdoc comments

merkle/src/tree.rs
```diff
@@ -0,0 +1,22 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0
+
+/// Trait for implementing general binary merkle tree
```
jspada, line 4: nice rustdoc!

merkle/src/tree_impl.rs
```diff
@@ -0,0 +1,192 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0
+
+use super::*;
+use std::marker::PhantomData;
+
+const DEGREE: usize = 2;
+
+/// Special complete binary merkle tree that is compatible with
```
jspada, line 9: Is compatibility important?

merkle/src/tree_impl.rs
jspada: Please confirm here ocaml does not used a duplicate of a sibling's value if it's odd.

merkle/src/tree_impl.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    fn count(&self) -> usize {
+        self.leafs.len()
+    }
+
+    fn root(&mut self) -> Option<Self::Hash> {
+        self.calculate_hash_if_needed(0).map(|(hash, _)| hash)
+    }
+
+    fn add_batch(&mut self, items: Vec<Self::Item>) {
```
jspada, line 116: Please add rustdoc and normal comments for all of this, even internal things.

merkle/src/tree_impl.rs
```diff
@@ -0,0 +1,192 @@
...
+    debug_assert!(index > 0);
+    (index - 1) / 2
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn calculate_depth_tests() {
```
jspada, line 182: Please add more unit tests.

merkle/tests/tree.rs
```diff
@@ -0,0 +1,136 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0
+
+// Test cases are generated from ocaml code
```
jspada, line 4: Nice job!

merkle/tests/tree.rs
jspada: Nice work @hanabi1224  - that's the right way forward to blackbox the hash function for now as you have. 👍🏼

jspada (changes requested):

jspada:
Awesome work!  Requested some changes, mostly just adding comments and rustdoc.

merkle/src/tree_impl.rs
hanabi1224: Compatibility is more referring to the fact that a mina merkle tree puts all leaf nodes at the same height/depth, while in [other implementations](https://crates.io/crates/merkle-cbt), it's not required, thus the tree can be more compact. And because mina uses a depth-aware hash algorithm, I think the latter will break compatibility.

merkle/src/tree_impl.rs
hanabi1224: Confirmed, also the even-odd case is covered by the unit test cases that are generated from OCaml code.

merkle/src/tree_impl.rs
hanabi1224: This is trait implementation, the generated rust doc will automatically inherit the doc from the Trait definition, any rust doc here will override the inherited doc.

To verify this, run `cargo doc --no-deps --release --all-features` and check generated doc under target/docs/mina_merkle

merkle/src/tree_impl.rs
hanabi1224: Will have more tests for real mina hasher and merger implementation. For now I will add a second merger that duplicates the odd leaf, code coverage report can be checked here [here](https://app.codecov.io/gh/ChainSafe/mina-rs/compare/123)

merkle/src/tree_impl.rs
hanabi1224: Also added a second TestMerger that duplicates the odd leaf

merkle/tests/tree_dup_odd_leaf_merger.rs
```diff
@@ -0,0 +1,137 @@
...
+        ) -> Option<Self::Hash> {
+            if let Some((left, _)) = items[0] {
+                if let Some((right, _)) = items[1] {
+                    Some(left + right)
+                } else {
+                    Some(*left * 2)
+                }
+            } else {
+                None
+            }
```
willemolding, line 52: A bit more compact and readable if you like

```suggestion
            match (items[0], items[1]) {
                (Some((left, _)), Some((right, _))) => Some(left+right),
                (Some((left, _)), None)) => Some(*left * 2),
                _ => None,
            }
```

hanabi1224: Fixed, thx!

merkle/tests/tree.rs
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    struct TestMerger {}
+
+    impl MerkleMerger<2> for TestMerger {
+        type Hash = i64;
+        fn merge(
+            items: [&Option<(Self::Hash, MerkleTreeNodeMetadata<2>)>; 2],
+        ) -> Option<Self::Hash> {
+            if let Some((left, _)) = items[0] {
```
willemolding, line 44: same here

hanabi1224: Fixed!

merkle/src/tree.rs
```diff
@@ -0,0 +1,22 @@
...
+    fn depth(&self) -> u32;
+    /// Number of leafs
+    fn count(&self) -> usize;
+    /// Root hash, lazy-evaluated
+    fn root(&mut self) -> Option<Self::Hash>;
+
+    /// Adds a new leaf
+    fn add(&mut self, item: Self::Item);
+    /// Adds a batch of leaves in the give order
+    fn add_batch(&mut self, items: Vec<Self::Item>);
```
willemolding, line 21: I think requiring a function to accept a vector isn't the best practice. Instead something like
```rust 
fn add_batch<I: IntoIterator<Self::Item>>(&mut self, items: I);
```
would be better. This would eliminate a bunch of conversions to vectors and let us pass arrays, slices and other things

willemolding (changes requested):
One minor change to the MerkleTree trait but otherwise looks great!

merkle/src/tree.rs
hanabi1224: Fixed

willemolding (approved):

crypto/src/hash/prefixes.rs
hanabi1224: This was a const fn to calculate predefined prefixes at compile time instead of runtime, so str or string types are not feasible, will rename the locals  (this fn is not introduced in this PR tho)

jspada (approved):

PR #128, opened 2022-01-19
minor refactors around consensus and use actual is_short_range API

creativcoder:
**Summary of changes**
- 

Closes 

### DIFF
--- a/consensus/src/checkpoint.rs
+++ b/consensus/src/checkpoint.rs
@@ -3,15 +3,12 @@
 
 use crate::common::{Chain, ProtocolStateChain};
 use crate::constants::SLOTS_PER_EPOCH;
+use crate::error::ConsensusError;
 use mina_crypto::base58::Base58Encodable;
 use mina_crypto::hash::{EpochSeed, StateHash};
 use mina_rs_base::protocol_state::ProtocolState;
 
-#[derive(Debug)]
-pub enum ConsensusErrTyp {
-    ConsensusInitFail,
-}
-pub fn init_checkpoints(genesis: &mut ProtocolState) -> Result<(), ConsensusErrTyp> {
+pub fn init_checkpoints(genesis: &mut ProtocolState) -> Result<(), ConsensusError> {
     genesis.body.consensus_state.staking_epoch_data.seed = EpochSeed::default();
     genesis
         .body
@@ -31,7 +28,7 @@ pub fn init_checkpoints(genesis: &mut ProtocolState) -> Result<(), ConsensusErrT
         .0 = 1;
     genesis.body.consensus_state.next_epoch_data.seed =
         EpochSeed::from_base58("2vaRh7FQ5wSzmpFReF9gcRKjv48CcJvHs25aqb3SSZiPgHQBy5Dt")
-            .map_err(|_| ConsensusErrTyp::ConsensusInitFail)?;
+            .map_err(|_| ConsensusError::ConsensusStateNotFound)?;
     genesis
         .body
         .consensus_state
@@ -39,7 +36,7 @@ pub fn init_checkpoints(genesis: &mut ProtocolState) -> Result<(), ConsensusErrT
         .start_checkpoint = StateHash::default();
     genesis.body.consensus_state.next_epoch_data.lock_checkpoint =
         StateHash::from_base58("3NLoKn22eMnyQ7rxh5pxB6vBA3XhSAhhrf7akdqS6HbAKD14Dh1d")
-            .map_err(|_| ConsensusErrTyp::ConsensusInitFail)?;
+            .map_err(|_| ConsensusError::ConsensusStateNotFound)?;
     genesis.body.consensus_state.next_epoch_data.epoch_length.0 = 2;
     Ok(())
 }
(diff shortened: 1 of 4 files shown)


consensus/src/checkpoint.rs
```diff
@@ -3,15 +3,12 @@
...
 use mina_crypto::base58::Base58Encodable;
 use mina_crypto::hash::{EpochSeed, StateHash};
 use mina_rs_base::protocol_state::ProtocolState;
 
-#[derive(Debug)]
-pub enum ConsensusErrTyp {
-    ConsensusInitFail,
-}
-pub fn init_checkpoints(genesis: &mut ProtocolState) -> Result<(), ConsensusErrTyp> {
+pub fn init_checkpoints(genesis: &mut ProtocolState) -> Result<(), ConsensusError> {
```
hanabi1224, line 11: How about creating a Checkpoint trait and making this a trait method? In general should we prefer OO way in our api design or non-OO way?

consensus/src/common.rs
```diff
@@ -117,11 +118,6 @@ pub trait Consensus {
     ) -> Result<&'a ProtocolStateChain, ConsensusError>;
 }
 
-// TODO: replace from checkpoint.rs
-fn is_short_range(_cs: &ProtocolStateChain) -> bool {
```
polarbeargo, line 121: It's conflict with spec https://github.com/MinaProtocol/mina/blob/32f529d8d8d712a44ee75be66061ce08cbdc8924/docs/specs/consensus/README.md#532-short-range-fork-check

polarbeargo (changes requested):

consensus/src/checkpoint.rs
creativcoder: Not sure how we could possibly enforce OO around every code, but definitely trait based APIs wherever possible (the rusty way). Yes am thinking of refactoring the checkpoint related parts, as part of this PR.

consensus/src/common.rs
creativcoder: yes, this has been replaced with the is_short_range function from `checkpoint.rs`

consensus/src/checkpoint.rs
creativcoder: I notice the `init_checkpoint` function is only being used in test cases at the moment, so i have moved it to under `tests/checkpoint.rs` within `mod tests {}`. There isn't really much to refactor in checkpoint.rs related to code design at the moment as we simply have a `is_short_range` function there. Based on how it's being used in `select_secure_chain` API, it feels having `is_short_range` as a free helper function is appropriate.

hanabi1224 (approved):

noot (approved):

consensus/tests/checkpoint.rs
```diff
@@ -3,11 +3,44 @@
...
+            .staking_epoch_data
+            .lock_checkpoint = StateHash::default();
+        genesis
+            .body
+            .consensus_state
+            .staking_epoch_data
+            .epoch_length
+            .0 = 1;
+        genesis.body.consensus_state.next_epoch_data.seed =
+            EpochSeed::from_base58("2vaRh7FQ5wSzmpFReF9gcRKjv48CcJvHs25aqb3SSZiPgHQBy5Dt")
```
noot, line 30: is this arbitrary?

consensus/src/checkpoint.rs
```diff
@@ -3,57 +3,18 @@
...
-    genesis.body.consensus_state.next_epoch_data.epoch_length.0 = 2;
-    Ok(())
-}
+use crate::error::ConsensusError;
 
 pub fn is_short_range(
     c0: &ProtocolStateChain,
     c1: &ProtocolStateChain,
-) -> Result<bool, ConsensusErrTyp> {
+) -> Result<bool, ConsensusError> {
```
polarbeargo, line 11: Awesome work on standout help defines `ConsensusError` 🥇 .

consensus/src/common.rs
polarbeargo: Great work on helping modify this part 👍 !

consensus/src/checkpoint.rs
```diff
@@ -3,57 +3,18 @@
...
-        .body
-        .consensus_state
-        .next_epoch_data
-        .start_checkpoint = StateHash::default();
-    genesis.body.consensus_state.next_epoch_data.lock_checkpoint =
-        StateHash::from_base58("3NLoKn22eMnyQ7rxh5pxB6vBA3XhSAhhrf7akdqS6HbAKD14Dh1d")
-            .map_err(|_| ConsensusErrTyp::ConsensusInitFail)?;
-    genesis.body.consensus_state.next_epoch_data.epoch_length.0 = 2;
-    Ok(())
-}
```
polarbeargo, line 45: Uncertain if we can totally ignore this lifecycle for checkpointing?

consensus/src/error.rs
```diff
@@ -13,6 +13,4 @@ pub enum ConsensusError {
     ConsensusStateNotFound,
```
polarbeargo, line 13: Beautiful work on helping define Consensus in a consistent way 🚀 !

consensus/tests/checkpoint.rs
creativcoder: No, it's from the spec: https://github.com/MinaProtocol/mina/blob/02dfc3ff0160ba3c1bbc732baa07502fe4312b04/docs/specs/consensus/README.md#531-genesis-checkpoints

consensus/src/checkpoint.rs
creativcoder: I could not follow the question, could you rephrase?

polarbeargo: I'm wondering if in Blockchain Mina protocol checkpoint has a life cycle as [Life Cycle of a Servlet](https://www.geeksforgeeks.org/life-cycle-of-a-servlet/) has to init(), resume(),  and destroy()? Since we have init_checkpoints() to init its states. If yes we will need to handle the life cycle with care. Probably will not just throw it into the test module.

creativcoder: No they are not related to servlet lifecycle methods. As far as i see it, the init_checkpoint function was just a helper function created to make writing test cases easier by helping stub the staking epoch and next epoch fields.

consensus/src/checkpoint.rs
hanabi1224: Sounds good, thx!

polarbeargo (approved):
Thank you for helping refactors, Excellent work 🚀

Issue #138, opened 2022-01-24
Refactor deserialization to a single crate of types with Into/From implementations for internal types


willemolding:
This refactor gives total separation between the type definitions which exist just to generate serde serialization impls and the types we will use internally to implement various algorithms.

This will make the core code much more readable and more resilient to changes in the existing Mina wire protocol. Keeping all of these types in a single crate will simplify the testing dependencies and allow us to ship a completed crate for any project that wants to consume the Mina wire types.

PR #150, opened 2022-02-01
Add network types crate and versioned crate

willemolding:
**Summary of changes**

I've split issue #138 into a number of PRs for ease of review. This one only adds things.

- Adds the `network-types` crates which collates all of the types used to generate the serde serialization code
- Copies some tests from the test-serialization crate to become the tests for the new crate. Propose removing the tests from the other crate in a future PR
- Adds the `versioned` crate  which supports the `network-types` crate and adds a more explicit method for versioning the types that does not require a proc macro. Types now carry their version info with them. This means if we do need to update a type to a later version the type-checker will be able to guide us through the update. This replaces the `wire-type` crate which should also be removed in a future PR.


polarbeargo (approved):
An astonished, amazing, and elegant work on resolving our protocol version problem 🚀

protocol/network-types/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+edition = "2021"
+license = "Apache-2.0"
+
+[dependencies]
+versioned = {path = "../versioned"}
+serde = {version = "1", features = ["derive"]}
+mina-crypto = {path = "../../crypto"} # TODO: Try and remove this dep
+
+[dev-dependencies]
+wasm-bindgen-test = "0.3"
```
hanabi1224, line 15: Where is `wasm-bindgen-test` used? Plz note that wasm tests in new crates need extra setup in CI workflow yaml

protocol/network-types/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+authors = ["ChainSafe Systems <info@chainsafe.io>"]
+name = "mina-network-types"
```
hanabi1224, line 3: The serialized bytes can also be used for storage, maybe `serializable-types` is better than `network-types`?

hanabi1224 (approved):
Nice work! Only some minor comments on naming

PR #151, opened 2022-02-03
Cleanup consensus crate and initial short range fork tests

creativcoder:
**Summary of changes**
- Cleans up consensus crate and relocates a few items to common.rs.
- Moves the consensus constants under the Consensus trait as const, so APIs don't need to pass them around.
- Started with adding basic tests for short range fork, with data from live blocks from mainnet.

Closes 


consensus/src/common.rs
```diff
@@ -3,15 +3,46 @@
...
+    pub sub_windows_per_window: Length,
+    /// Number of slots before minimum density is used in chain selection
+    pub grace_period_end: Length,
+}
 
-use crate::checkpoint::is_short_range;
-use crate::density::{relative_min_window_density, ConsensusConstants};
-use crate::error::ConsensusError;
+impl ConsensusConstants {
+    pub const fn from_genesis() -> Self {
```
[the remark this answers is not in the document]
creativcoder: Right, the ocaml source derives these out of ProtocolConstants in addition to whether the env is mainnet or devnet. I'll refactor this to cover both mainnet and devnet.

consensus/src/common.rs
```diff
@@ -106,6 +137,7 @@ impl Chain<ProtocolState> for ProtocolStateChain {
 
 pub trait Consensus {
     type Chain;
+    const CONSTANTS: ConsensusConstants;
```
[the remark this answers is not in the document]
creativcoder: Yeah, i just skimmed the ocaml source and it makes sense to move this from here and allow it to be loadable from a file. Adding a constants method sounds about right at the moment.

consensus/src/common.rs
```diff
@@ -116,17 +148,20 @@ pub trait Consensus {
...
+    fn is_short_range(&self, candidate: &ProtocolStateChain) -> Result<bool, ConsensusError>;
 }
 
 impl Consensus for ProtocolStateChain {
     type Chain = ProtocolStateChain;
+    const CONSTANTS: ConsensusConstants = ConsensusConstants::from_genesis();
     fn select_secure_chain<'a>(
         &'a self,
         candidates: &'a [Self::Chain],
         constants: &ConsensusConstants,
```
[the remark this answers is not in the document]
creativcoder: Ah good catch, this shouldn't be passed to `select_secure_chain` as we expect the constants to be present on the implementer.

consensus/tests/checkpoint.rs
```diff
@@ -1,123 +1,222 @@
...
+    #[should_panic]
+    fn short_range_fails_when_chains_are_one_epoch_apart_but_beyond_seed_update_range() {
+        let mut chain_a = genesis_consensus_state();
+        let a = &mut chain_a.0[0];
+
+        // A block at height 5076
+        // block in json: https://storage.googleapis.com/mina_network_blocfk_data/mainnet-3NLC8CV9kZYFkXnUipJzkkHvT9RmsttSUNpwfwqwWCfbP9bQmwNJ.json
+        let mut chain_b = genesis_consensus_state();
+        let b = &mut chain_b.0[0];
+        b.body.consensus_state.epoch_count = Length(1);
```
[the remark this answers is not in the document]
creativcoder: I believe we'll be initializing them this way in only tests. I think it's okay at the moment.

creativcoder:
All code review comments have been addressed

willemolding (approved):

consensus/src/common.rs
```diff
@@ -242,7 +244,77 @@ impl Consensus for ProtocolStateChain {
...
+                    0,
+                ),
+                self.config().sub_windows_per_window.0,
+            );
+            // initialize projected window based off of chain_a
+            let mut projected_window = tip_state.sub_window_densities.clone();
+
+            // relative sub window
+            let mut rel_sub_window =
+                tip_state.curr_global_slot.slot_number.0 % self.config().sub_windows_per_window.0;
```
jspada, line 299: I think this should be

```rust
tip_state.curr_global_slot.slot_number.0/self.config().slots_per_sub_window.0 % self.config().sub_windows_per_window.0;
```

consensus/src/common.rs
```diff
@@ -242,7 +244,77 @@ impl Consensus for ProtocolStateChain {
...
+
+            // ring shift
+            while shift_count > 0 {
+                rel_sub_window = (rel_sub_window + 1) % self.config().sub_windows_per_window.0;
+                projected_window[rel_sub_window as usize] = Length(0);
```
jspada, line 304: I know this matches OCaml implementation, but calling this a `Length` is confusing.  It would be nicer to create a type called `Density` or to keep it simple, call it whatever it is.. e.g. `u32` or `u64`


Actually... on second thought, it probably ties back to serialisation and therefore would be a pain to change it and make things confusing.

So maybe because we want this to map to real mina  pub/network protocol types it would be too confusing to change it like I suggested earlier.

Unless the serialisation layout stuff had a way to map things to different names automatically.. but again, sounds complicated.

Thus, I retract my previous uninformed suggestion.  Instead maybe just put a comment next to it like `// really a density` or maybe you can just create a local alias!

consensus/src/common.rs
```diff
@@ -242,7 +244,77 @@ impl Consensus for ProtocolStateChain {
...
+        };
+
+        // compute projected window density
+        let projected_window_density = projected_window.iter().map(|s| s.0).sum();
+
+        // compute minimum window density
+        Ok(min(
+            tip_state.min_window_density.0,
+            projected_window_density,
+        ))
```
jspada, line 318: Very nice implementation!

consensus/tests/density.rs
```diff
@@ -0,0 +1,141 @@
...
+
+        let mut chain_1000_state = ProtocolState::default();
+        chain_1000_state.body.consensus_state = chain_1000.clone();
+        let mut chain_1000 = ProtocolStateChain::default();
+        chain_1000.push(chain_1000_state).unwrap();
+
+        let a = genesis.relative_min_window_density(&chain_1000);
+        assert_eq!(a.unwrap(), 77);
+    }
+}
```
jspada, line 141: Looks great, if you have test cases you can extract from live data, I think you should add some more to exercise more cases and seek out annoying edge-cases.

consensus/tests/checkpoint.rs
```diff
@@ -113,5 +123,100 @@ mod tests {
...
+
+        b.body.consensus_state.staking_epoch_data.lock_checkpoint =
+            StateHash::from_base58("3NKW9MqYePrfK48ZAn3iu7iDDc17wZkTmzo5tHQ96bvG1gUfG9Dv").unwrap();
+        b.body.consensus_state.staking_epoch_data.start_checkpoint =
+            StateHash::from_base58("3NKcySpVjMsrLpCHYvzmEvap4HUhSdws8HU7fv5cL3wgamYVzQts").unwrap();
+        b.body.consensus_state.curr_global_slot = GlobalSlot::default();
+        b.body.consensus_state.curr_global_slot.slot_number = GlobalSlotNumber(1577);
+        b.body.consensus_state.curr_global_slot.slots_per_epoch = Length(7140);
+
+        assert!(!chain_a.is_short_range(&chain_b).unwrap());
```
jspada, line 218: Great stuff

jspada (approved):
Great work - approving now for simplicity, but there's a few feedbacks to check and potentially address.

consensus/src/common.rs
```diff
@@ -3,15 +3,49 @@
...
+    /// Timestamp of genesis block in unixtime
+    pub genesis_state_timestamp: BlockTime,
+    /// Sub windows within a window
+    pub sub_windows_per_window: Length,
+    /// Number of slots before minimum density is used in chain selection
+    pub grace_period_end: Length,
+}
+
+impl ConsensusConstants {
+    pub fn mainnet() -> Self {
```
hanabi1224, line 33: how about making this a const fn?

consensus/src/common.rs
```diff
@@ -3,15 +3,49 @@
...
+            slots_per_epoch: Length(7140),
+            slots_per_sub_window: Length(7),
+            delta: Length(0),
+            genesis_state_timestamp: BlockTime(1615939200000),
+            sub_windows_per_window: Length(11),
+            grace_period_end: Length(1440),
+        }
+    }
+
+    pub fn devnet() -> Self {
```
hanabi1224, line 45: Ditto

consensus/src/common.rs
```diff
@@ -137,45 +140,44 @@ impl Chain<ProtocolState> for ProtocolStateChain {
...
-                    .consensus_state()
-                    .ok_or(ConsensusError::ConsensusStateNotFound)?;
-                let candidate_state = c
-                    .consensus_state()
-                    .ok_or(ConsensusError::ConsensusStateNotFound)?;
-                let tip_density =
-                    relative_min_window_density(tip_state, candidate_state, &Self::CONSTANTS)?;
-                let candidate_density =
-                    relative_min_window_density(candidate_state, tip_state, constants)?;
+                let tip_density = self.relative_min_window_density(c)?;
```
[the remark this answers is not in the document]
creativcoder: done

consensus/src/common.rs
```diff
@@ -228,7 +230,7 @@ impl Consensus for ProtocolStateChain {
 
         let check = |s1: &ConsensusState, s2: &ConsensusState, s2_epoch_slot: Option<u32>| {
             if s1.epoch_count.0 == s2.epoch_count.0 + 1
-                && s2_epoch_slot >= Some(Self::CONSTANTS.slots_per_epoch.0 * 2 / 3)
+                && s2_epoch_slot >= Some(self.config().slots_per_epoch.0 * 2 / 3)
```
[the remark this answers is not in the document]
creativcoder: done

consensus/src/common.rs
```diff
@@ -242,7 +244,77 @@ impl Consensus for ProtocolStateChain {
...
+        }
+    }
+
+    fn config(&self) -> ConsensusConstants {
+        ConsensusConstants::mainnet()
+    }
+
+    /// Computes the relative minimum window density of the given chains.
+    /// The minimum density value is used in the case of a long range fork
+    /// and the chain with the higher minimum window density is chosen as the canonical chain.
```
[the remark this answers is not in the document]
creativcoder: done

consensus/src/common.rs
creativcoder: ah good catch! thanks

consensus/src/common.rs
creativcoder: Yeah, `Length` is somewhat confusing, but it's there because it had to match mina's serialization types. I've just added a comment saying `// density at rel_sub_window index`

consensus/src/common.rs
creativcoder: At this moment, it's not entirely clear how the configs will eventually be loaded. It might also be the case that it may come from a config file.

consensus/tests/density.rs
creativcoder: Noted, we'll take up that as a subsequent PR.

jspada:
Awesome work!

This actually implements both short-range and long-range fork handling, but the tickets looks like it just covers short-range.

Since you've already crossed the rubicon, I think you should finish it off by implementing the long-range unit tests before you forget it all and debugging is too hard later on.

PR #159, opened 2022-02-18
M2: Setup basic peer initialization stubs: Initialize consensus step

### LINKED ISSUE #89: M2 - Implement Node and SyncManager abstractions
**Issue summary**

* Implement a `Node` (a participant in the peer to peer network) that initializes its state from genesis or the local best tip.
* Implement a `SyncManager` responsible for syncing blocks received from the network. This will eventually contain routines for short range fork and long range fork

creativcoder:
**Summary of changes**
- Creates a Peer entity that is responsible for connecting to peers and initalizing the node bootstrap sequence.
- Eventually calls the select chain API as part of its lifecycle method. [TODO]

Notes:
- We don't intend to implement the transition frontier in the peer as of now as it's still in flux.

Ref: https://github.com/MinaProtocol/mina/blob/02dfc3ff0160ba3c1bbc732baa07502fe4312b04/docs/specs/consensus/README.md#61-initialize-consensus

Closes #89

### DIFF
--- a/consensus/src/common.rs
+++ b/consensus/src/common.rs
@@ -3,11 +3,14 @@
 
 //! Implements common APIs for the blockchain in the context of consensus.
 
+use std::collections::HashMap;
+
 use hex::ToHex;
 use mina_crypto::hash::{Hashable, StateHash};
 use mina_rs_base::consensus_state::ConsensusState;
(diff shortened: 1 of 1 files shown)


creativcoder:
This is still being considered if we need it as part of m2, so closing this draft PR.

PR #162, opened 2022-02-24
Trait for constructing ROInput from field(s)

hanabi1224:
**Summary of changes**
This is a prerequisite of Merkle hasher that calculates hash value from Account data structure.

- Add a trait that facilitates constructing ROInput.
- Account field types should implement this trait
- Account type should implement mina_signer::Hashable
- the functionality this trait provides is equivalent to `to_input` methods that each Account field implements.
- Unit tests are generated from OCaml code

Closes 


crypto/src/hash/traits.rs
```diff
@@ -36,6 +37,12 @@ where
     }
 }
 
+/// Trait that converts itself into either field(s) or bits(s)
+/// for constructing the given [ROInput]
+pub trait RandomOraclePartialInput {
+    fn add_self_to(&self, input: &mut ROInput);
```
willemolding, line 43: The construction of this trait feels a bit strange to me. Since the `ROInput` is the thing being mutated it would be much more natural to have the trait implemented on that and be generic over what is being added. e.g. following the example of Vec::push

```rust
pub trait RandomOraclePartialInput<T> {
    fn append(&mut self, value: T);
}
```

An alternative would be to derive the[ `core::ops::Add`](https://doc.rust-lang.org/std/ops/trait.Add.html) trait which wouldn't mutate the ROInput but could return a new one with the value added.

I think either of these would make more sense and make it easier to collect the implementations in one place rather than scattering them around the codebase to where the things to be added are located.

willemolding (commented):
Unsure about the trait introduced in this PR

crypto/src/hash/traits.rs
hanabi1224: That's a great suggestion! Fixed as suggested, thanks!

willemolding (approved):
This makes way more sense to me

crypto/src/signature.rs
```diff
@@ -20,6 +24,47 @@ pub struct CompressedCurvePoint {
     pub is_odd: bool,
 }
 
+impl CompressedCurvePoint {
```
jspada, line 27: We do not want this re-implemented.  Instead we should provide some safe bridging code between proof-systems structs/types and those provided by Mina/ocaml deserialization output

crypto/src/hash/traits.rs
```diff
@@ -36,6 +36,12 @@ where
     }
 }
 
+/// Trait that converts given generic value into either field(s) or bits(s)
+/// and append to mutable [mina_signer::ROInput], this trait should be implemented for [mina_signer::ROInput]
+pub trait RandomOraclePartialInput<T> {
```
jspada, line 41: It is not clear what this trait is meant to provide that ROInput doesn't already.  Can you please explain?

With respect to the concept of "partial" random oracle input, there should be no partial implementation of the interface-- anything hashed of signed requires both `to_roinput` and `domain_string`.  If we find something in the ocaml that doesn't have both, then it is probably a bug.

Again you should hold off on this implementation until we update our Signable and Hashable interfaces for you.

hanabi1224: In the ocaml code, every field type of the `Account` data structure has `to_input` function, and the `to_input` of `Account` simply merges `to_input` result of every field, this it flexible to allow an arbitrary combination of the field types into a new data structure (if needed) whose `to_input` method can be easily implemented. 

An alternative solution could be, implementing `Hashable` for each field type, and then ues another API to merge multiple `ROInput`s into one. Does that make sense?

hanabi1224:
Superceded by https://github.com/ChainSafe/mina-rs/pull/198

crypto/src/hash/traits.rs
jspada: Done.

PR #167, opened 2022-02-28
Additional test cases for select longer chain and select secure chain

### LINKED ISSUE #165: [M2] Select secure chain test cases
**Issue summary**

We have the selectSecureChain API implemented. It would be great to have it tested against block data from mainnet.

Relevant spec section: https://github.com/MinaProtocol/mina/blob/02dfc3ff0160ba3c1bbc732baa07502fe4312b04/docs/specs/consensus/README.md#62-select-chain

sudo-shashank:
**Summary of changes**
- Added select longer chain test cases
- Added select secure chain test cases

Closes #165

### DIFF
--- a/consensus/tests/common.rs
+++ b/consensus/tests/common.rs
@@ -256,4 +256,231 @@ mod tests {
             VrfOutputTruncated::from("kKr83LYd7DyFupRAPh5Dh9eWM1teSEs5VjU4XId2DgA=")
         );
     }
+
+    #[test]
+    #[wasm_bindgen_test]
+    fn test_select_secure_chain_short_range_fork() {
+        let mut chain_a = ProtocolStateChain::default();
+        let mut consensus_state = ConsensusState::default();
+        consensus_state.epoch_count = Length(5);
+        consensus_state.blockchain_length = Length(11);
+        let mut prot_state = ProtocolState::default();
+        prot_state.body.consensus_state = consensus_state;
+        chain_a.push(prot_state).unwrap();
+
+        let mut chain_b = ProtocolStateChain::default();
+        let mut consensus_state = ConsensusState::default();
+        consensus_state.epoch_count = Length(5);
+        consensus_state.blockchain_length = Length(10);
+        let mut prot_state = ProtocolState::default();
+        prot_state.body.consensus_state = consensus_state;
+        chain_b.push(prot_state).unwrap();
+
+        let mut chains = vec![];
+        chains.push(chain_b);
+        let select_result = chain_a.select_secure_chain(&chains).unwrap();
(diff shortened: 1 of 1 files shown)


hanabi1224 (approved):

consensus/tests/common.rs
```diff
@@ -194,4 +194,294 @@ mod tests {
         assert_eq!(a.body.consensus_state.min_window_density, Length(43));
         assert_eq!(a.body.consensus_state.sub_window_densities, densities);
     }
+
+    #[test]
+    #[wasm_bindgen_test]
+    fn test_longer_chain_with_diff_blockchain_length() {
+        let mut chain_a = ProtocolStateChain::default();
+        let mut consensus_state = ConsensusState::default();
+        consensus_state.blockchain_length = Length(113267);
```
creativcoder, line 203: Are these mocked values or taken from mainnet?

consensus/tests/common.rs
```diff
@@ -194,4 +194,294 @@ mod tests {
...
+        assert_eq!(
+            result_state.body.consensus_state.blockchain_length,
+            Length(11)
+        );
+    }
+
+    #[test]
+    #[wasm_bindgen_test]
+    //Long range fork with greater relative min window density
+    fn test_select_secure_chain_long_range_fork_1() {
```
creativcoder, line 290: can we rename the test function and others to something more descriptive, just like how we have for the first test case: `test_longer_chain_with_diff_blockchain_length`

consensus/tests/common.rs
sudo-shashank: from mainnet but selected randomly

consensus/tests/common.rs
sudo-shashank: yes, will do

consensus/tests/common.rs
```diff
@@ -194,4 +194,294 @@ mod tests {
...
+        let constants = ConsensusConstants::from_genesis();
+        let mut chain_a = ProtocolStateChain::default();
+        let mut consensus_state = ConsensusState::default();
+        consensus_state.epoch_count = Length(23);
+        consensus_state.min_window_density = Length(77);
+        consensus_state.sub_window_densities = vec![
+            1.into(),
```
willemolding, line 297: All occurences of these will need to be replaced by 
```suggestion
            Length(1),
```
Sorry for the breaking change..

willemolding: Might make sense to refactor this out as a `const` or `const fn` since it appears so often

consensus/tests/common.rs
sudo-shashank: done

willemolding (approved):

PR #172, opened 2022-03-08
Added adversary and few more test cases for select secure chain method

### LINKED ISSUE #165: [M2] Select secure chain test cases

sudo-shashank:
**Summary of changes**
- Added comments for test cases 
- Used mainnet block data for some test cases
- Added following adversary test cases:
1.Adding extra sub-windows
2.Adding impossibly high sub-window densities
3.Having too small number of sub-windows
- Added error handling in relative_min_window_density
 
Closes #165

### DIFF
--- a/consensus/src/common.rs
+++ b/consensus/src/common.rs
@@ -6,8 +6,8 @@
 //!
 
 use crate::error::ConsensusError;
-use blake2::Blake2bVar;
 use blake2::digest::Update;
+use blake2::Blake2bVar;
 use lockfree_object_pool::SpinLockObjectPool;
 use mina_rs_base::consensus_state::ConsensusState;
 use mina_rs_base::global_slot::GlobalSlot;
@@ -135,7 +135,15 @@ impl Chain<ProtocolState> for ProtocolStateChain {
     fn last_vrf_hash(&self) -> Result<String, ConsensusError> {
         use blake2::digest::VariableOutput;
         let mut hasher = Blake2bVar::new(32).unwrap();
-        hasher.update(self.top().ok_or(ConsensusError::TopBlockNotFound)?.body.consensus_state.last_vrf_output.0.as_slice());
+        hasher.update(
+            self.top()
+                .ok_or(ConsensusError::TopBlockNotFound)?
+                .body
+                .consensus_state
+                .last_vrf_output
+                .0
+                .as_slice(),
+        );
         let hash = hasher.finalize_boxed();
         Ok(hex::encode(hash))
     }
--- a/consensus/src/common.rs
+++ b/consensus/src/common.rs
@@ -85,7 +85,7 @@ where
     fn length(&self) -> usize;
     /// This function returns the hex digest of the hash of the last VRF output
     ///  of a given chain. The input is a chain C and the output is the hash digest.
-    fn last_vrf_hash(&self) -> Result<String, ConsensusError>;
+    fn last_vrf_hash_digest(&self) -> Result<String, ConsensusError>;
     /// This function returns hash of the top block's protocol state for a given chain.
     /// The input is a chain C and the output is the hash.
     fn state_hash(&self) -> Option<Fp>;
@@ -133,7 +133,7 @@ impl Chain<ProtocolState> for ProtocolStateChain {
         self.0.len()
     }
 
-    fn last_vrf_hash(&self) -> Result<String, ConsensusError> {
+    fn last_vrf_hash_digest(&self) -> Result<String, ConsensusError> {
         let output_size = 32; // TODO: do we need this configurable?
         let mut hasher = Blake2bVar::new(output_size)
             .map_err(|_| ConsensusError::InvalidBlake2bOutputSize(output_size))?;
@@ -247,7 +247,10 @@ impl Consensus for ProtocolStateChain {
             return Ok(candidate);
         } else if top_state.blockchain_length == candidate_state.blockchain_length {
             // tiebreak logic
-            match candidate.last_vrf_hash()?.cmp(&self.last_vrf_hash()?) {
+            match candidate
+                .last_vrf_hash_digest()?
+                .cmp(&self.last_vrf_hash_digest()?)
+            {
                 std::cmp::Ordering::Greater => return Ok(candidate),
                 std::cmp::Ordering::Equal => {
                     if candidate.state_hash() > self.state_hash() {
--- a/consensus/src/common.rs
+++ b/consensus/src/common.rs
@@ -15,8 +15,8 @@ use mina_rs_base::types::{BlockTime, Length};
 use once_cell::sync::OnceCell;
 use proof_systems::mina_hasher::{create_kimchi, Fp, Hasher, PoseidonHasherKimchi};
 
-/// Type that defines constant values for mina consensus calculation
 // TODO: derive from protocol constants
+/// Constants used for the conensus
 pub struct ConsensusConstants {
     /// Point of finality (number of confirmations)
     pub k: Length,
@@ -55,7 +55,7 @@ impl ConsensusConstants {
 }
 
 /// A chain of [ProtocolState]
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Default, PartialEq, Clone)]
 // TODO: replace vec element with ExternalTransition
 pub struct ProtocolStateChain(pub Vec<ProtocolState>);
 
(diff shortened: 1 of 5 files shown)


consensus/src/common.rs
```diff
@@ -301,7 +301,10 @@ impl Consensus for ProtocolStateChain {
             // ring shift
             while shift_count > 0 {
                 rel_sub_window = (rel_sub_window + 1) % self.config().sub_windows_per_window.0;
-                projected_window[rel_sub_window as usize] = Length(0);
+                match projected_window.get_mut(rel_sub_window as usize) {
```
creativcoder, line 304: Was this triggered because of the adversarial test cases?

consensus/tests/common.rs
```diff
@@ -200,7 +200,9 @@ mod tests {
 
     #[test]
     #[wasm_bindgen_test]
-    fn test_longer_chain_with_diff_blockchain_length() {
+    //Current chain: Chain A (Long Chain)
```
creativcoder, line 203: nit: space after `//`

consensus/src/common.rs
sudo-shashank: yes

consensus/tests/common.rs
sudo-shashank: sure i'll take care of this at other places too

consensus/tests/common.rs
```diff
@@ -480,7 +727,79 @@ mod tests {
...
+
+        let mut chain_b = ProtocolStateChain::default();
+        let mut consensus_state = ConsensusState::default();
+        consensus_state.epoch_count = Length(15);
+        consensus_state.min_window_density = Length(33);
+        consensus_state.sub_window_densities = vec![
+            Length(6),
+            Length(1),
+            Length(3),
+            // Below sub window densities are missing, resulting in index out of bound error
```
creativcoder, line 777: Are these missing from the block json repr or it's a delibrate removal to represent adversarial test case?

sudo-shashank: delibrate removal to represent adversarial test case

creativcoder (approved):
Nice work on these! Left a few comments

hanabi1224 (approved):

consensus/tests/common.rs
```diff
@@ -255,10 +289,57 @@ mod tests {
...
+
+        let result_state = select_result.0.get(0).unwrap();
+        assert_eq!(
+            result_state.body.consensus_state.last_vrf_output,
+            VrfOutputTruncated::from("kKr83LYd7DyFupRAPh5Dh9eWM1teSEs5VjU4XId2DgA=")
+        );
+        assert_eq!(result_state, chain_a.0.get(0).unwrap());
     }
 
+    /* TODO: Blocked due to poseidon hash implementation
```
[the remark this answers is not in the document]
sudo-shashank: sure

consensus/tests/common.rs
```diff
@@ -412,19 +655,24 @@ mod tests {
         chains.push(chain_b);
         let select_result = chain_a.select_secure_chain(&chains).unwrap();
         let result_state = select_result.0.get(0).unwrap();
-        assert_eq!(
+        assert_ne!(
             result_state.body.consensus_state.blockchain_length,
-            Length(77748)
+            Length(113267)
         );
+        assert_ne!(result_state, chain_a.0.get(0).unwrap());
```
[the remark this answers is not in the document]
sudo-shashank: yes right `chain_b` is the adversarial chain and result should actually stay on `chain a` 
but verification of these in-variants is done elsewhere in proof-checking i guess and not in `select_secure_chain` api
so these will fail and select incorrect chain as of now, hence i used `assert_ne!`

consensus/tests/common.rs
```diff
@@ -443,44 +691,114 @@ mod tests {
...
+        let mut prot_state = ProtocolState::default();
+        prot_state.body.consensus_state = consensus_state;
+        chain_b.push(prot_state).unwrap();
+
+        let mut chains = vec![];
+        chains.push(chain_b);
+        let select_result = chain_a.select_secure_chain(&chains).unwrap();
+        let result_state = select_result.0.get(0).unwrap();
+
+        assert_ne!(result_state, chain_a.0.get(0).unwrap());
```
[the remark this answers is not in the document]
sudo-shashank: https://github.com/ChainSafe/mina-rs/pull/172#discussion_r837164205

consensus/src/common.rs
```diff
@@ -176,6 +176,20 @@ impl Consensus for ProtocolStateChain {
...
+                    .ok_or(ConsensusError::ConsensusStateNotFound)?;
+
+                if tip_state.sub_window_densities.len() != 11 {
+                    return Ok(candidate);
+                } else if candidate_state.sub_window_densities.len() != 11 {
+                    return Ok(self);
+                }
+
                 let tip_density = self.relative_min_window_density(candidate)?;
                 let candidate_density = candidate.relative_min_window_density(self)?;
```
sudo-shashank, line 194: Since we are already checking the window densities len here, the error won't be propagated to the caller from `relative_min_window_density`
code suggestion:
> let tip_density = match self.relative_min_window_density(candidate) {
Ok(d) => Ok(candidate),
Err(err) => // Maybe log this error
}
let candidate_density = match candidate.relative_min_window_density(self) {
Ok(d) => Ok(candidate),
Err(err) => // Maybe log this error
}

creativcoder: Yeah, the idea is to not even consider computing the relative density if the window length is not of expected length.

sudo-shashank: then can we remove len check from `relative_min_window_density`?

```
if tip_state.sub_window_densities.len() != 11 || chain_b.sub_window_densities.len() != 11 {
            return Err(ConsensusError::InvalidSubWindowDensityLen);
        }
```

creativcoder: Yeah, this will be removed

consensus/src/common.rs
```diff
@@ -176,6 +176,21 @@ impl Consensus for ProtocolStateChain {
...
+                    .consensus_state()
+                    .ok_or(ConsensusError::ConsensusStateNotFound)?;
+
+                let sub_windows_per_window = self.config().sub_windows_per_window.0 as usize;
+                if tip_state.sub_window_densities.len() != sub_windows_per_window {
+                    return Ok(candidate);
+                } else if candidate_state.sub_window_densities.len() != sub_windows_per_window {
+                    return Ok(self);
+                }
+
```
sudo-shashank, line 193: @creativcoder since this is a part which impacts `relative_min_window_density`, adding this check at the starting of `relative_min_window_density` could be a better idea 
also as there are additional cases which we need to handle in this method, like extremely high sub window density
It makes more sense to me in adding it inside `relative_min_window_density`

creativcoder: I had it there within relative_min_window_density before, but relative_min_window_density's returns a u32, whereas we need to return the chain when the lengths are not ideal. As a result, had to move it above.

consensus/tests/common.rs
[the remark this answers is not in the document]
creativcoder: Question: As per consensus protocol, what's the max value a subwindow density can have?

consensus/src/common.rs
```diff
@@ -131,14 +133,21 @@ impl Chain<ProtocolState> for ProtocolStateChain {
...
 
-    fn last_vrf_hash(&self) -> Option<Fp> {
-        static HASHER_POOL: OnceCell<SpinLockObjectPool<PoseidonHasherKimchi<VrfOutputTruncated>>> =
-            OnceCell::new();
-        let pool =
-            HASHER_POOL.get_or_init(|| SpinLockObjectPool::new(|| create_kimchi(()), |_| ()));
-        let mut hasher = pool.pull();
-        self.top()
-            .map(|s| hasher.hash(&s.body.consensus_state.last_vrf_output))
+    fn last_vrf_hash(&self) -> Result<String, ConsensusError> {
```
willemolding, line 136: I don't think we should ever be using the `String` type for hashes. We have loads of different more informative hash types to use

willemolding (approved):

consensus/src/common.rs
creativcoder: Just following the spec here: https://github.com/MinaProtocol/mina/blob/compatible/docs/specs/consensus/README.md#516-hashlastvrf

jspada (approved):
All issues have been addressed.  Looks good to me!

PR #184, opened 2022-03-18
User command helpers

willemolding:
# User Command helpers

User commands are the Mina term for transactions, essentially ways to mutate the state. This PR makes it easier to construct these commands

**Summary of changes**
- Refactor user commands to their own module
- Adds doc comments
- Adds a builder to make producing signed balance transfer commands much easier
- Adds tests for builder


hanabi1224 (approved):

base/src/user_commands/memo.rs
```diff
@@ -0,0 +1,52 @@
...
+}
+
+impl TryFrom<String> for SignedCommandMemo {
+    type Error = SignedCommandMemoError;
+    fn try_from(s: String) -> Result<Self, Self::Error> {
+        Self::try_from(s.as_str())
+    }
+}
+
+// TODO impl Into<String> for SignedCommandMemo
```
creativcoder, line 44: Shouldn't it be `From<String> for SignedCommandMemo`?

base/src/user_commands/memo.rs
```diff
@@ -0,0 +1,52 @@
...
+pub struct SignedCommandMemo(pub Vec<u8>);
+
+impl TryFrom<&str> for SignedCommandMemo {
+    type Error = SignedCommandMemoError;
+    fn try_from(s: &str) -> Result<Self, Self::Error> {
+        const DIGEST_LEN: usize = 32;
+        const MAX_INPUT_STRING_LENGTH: usize = DIGEST_LEN;
+        const MEMO_LEN: usize = DIGEST_LEN + 2;
+        const TAG_INDEX: usize = 0;
+        // const DIGEST_TAG: u8 = 0;
```
creativcoder, line 21: can we remove this?

creativcoder (approved):

base/src/user_commands/memo.rs
willemolding: Actually it should probably `ToString for SignedCommandMemo`. `TryFrom<String>` is already implemented above

PR #186, opened 2022-03-22
M5 - Setup protobuf as msg protocol for js-wasm interop

hanabi1224:
**Summary of changes**

This PR introduces protobuf as js-wasm interop message protocol, doing automatic codegen from the same `.proto` files to avoid defining arbitrary json data structures on both sides

- Define js-wasm interop messages in `.proto` files
- Auto-codegen during build for `javascript`
- Auto-codegen during build for `rust`
- Update `license header checker` script for `javascript` and `typescript`

Closes  https://github.com/ChainSafe/mina-rs/issues/187


apps/wasm/mvp/web/pb/messages.js
```diff
@@ -0,0 +1,332 @@
+/*eslint-disable*/
```
willemolding, line 1: This is a generated file right? I think we should include a header warning people that it is generated and should not be edited manually.

We could also consider giving it a different file extension e.g. messages.autogen.js

apps/wasm/mvp/src/pb/responses.rs
```diff
@@ -0,0 +1,236 @@
+// This file is generated by rust-protobuf 2.27.1. Do not edit
```
willemolding, line 1: Oh yeah exactly like this one :rofl: but for .js / .ts files as well

apps/wasm/mvp/proto/messages.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+
+message PeerStatus
```
willemolding, line 3: Doc comments for the interop messages would be sweet

scripts/add_license.sh
```diff
@@ -1,11 +1,12 @@
 #!/bin/bash
 
 PAT_APA="^// Copyright 2020 ChainSafe Systems // SPDX-License-Identifier: Apache-2.0$"
+PAT_CODEGEN="^// This file is generated by "
 
 valid=true
 for file in $(find . -type f -not -path "./target/*" -not -path "./blockchain/beacon/src/drand_api/*" -not -path "./ipld/graphsync/src/message/proto/message.rs" | egrep '\.(rs)$'); do
```
willemolding, line 7: I think js files need the license header as well, we can aso remove the exceptions they are from Forest

willemolding (approved):

scripts/add_license.sh
hanabi1224: Good suggestion, fixed

apps/wasm/mvp/web/pb/messages.js
hanabi1224: Added the header in the custom codegen script since the codegen tool itself does not provide such an option.

> giving it a different file extension 

That will make the import a little ugly like `import { PeerStatus } from "./pb/messages.autogen";`, I guess that can be implied by the `/pb/` prefix

apps/wasm/mvp/proto/messages.proto
hanabi1224: Fixed

apps/wasm/mvp/src/utils.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0
+
+use js_sys::Uint8Array;
+use wasm_bindgen::JsValue;
+
+pub fn pb_to_u8a<T: protobuf::Message>(msg: &T) -> anyhow::Result<Uint8Array> {
```
creativcoder, line 7: does a more expanded name like: `proto_msg_to_u8array` sound good?

apps/wasm/mvp/src/lib.rs
```diff
@@ -17,56 +20,53 @@ use libp2p::{
...
 };
 use mina_network::p2p::{TransportBuilder, MAINNET_CONFIG};
+use pb::{requests::ConnectRequest, responses::CommonResponse};
 use std::{io, time::Duration};
-use wasm_bindgen::prelude::*;
+use utils::*;
 
 const TIMEOUT_SECS: u64 = 60;
 
 static mut EVENT_EMITTER: Option<EventEmitter> = None;
```
creativcoder, line 29: Instead of trying to initialize this using unsafe blocks ourselves, it'd be good if we can use something like once_cell: https://docs.rs/once_cell/latest/once_cell/ Thoughts?

hanabi1224: The js ffi type does not meet the generic constraint of `OnceCell`

```
`*mut u8` cannot be shared between threads safely
within `js_ffi::EventEmitter`, the trait `Sync` is not implemented for `*mut u8`
required because of the requirements on the impl of `Sync` for `once_cell::imp::OnceCell<js_ffi::EventEmitter>`
shared static variables must have a type that implements `Sync`
```

apps/wasm/mvp/src/utils.rs
hanabi1224: fixed

apps/wasm/mvp/src/lib.rs
hanabi1224: I've moved all `unsafe` code into a separate `event_emitter` mod and added `FIXME`.

creativcoder (approved):

PR #191, opened 2022-03-25
Added support for strong type deserialization of Polyvar

### LINKED ISSUE #185: M1 Support strongly typed deserialization of polyvars
**Task summary**

Bin-prot has two ways of representing enums. The first is as sum types which map perfectly to Rust enums. When encoded they include an index of the enum/sum variant followed by the data the variant contains.

Polyvars instead use a hash to indicate which variant. This is handy as it means you can re-arrange or add new variants without breaking existing serialized data. 

Currently our bin-prot crate supports deserializing polyvars using loose deserialization (with a layout file). This is able to persist the variant hashes. What we need is a way to have Rust enums be [de]serializable as polyvars. We need to copy the hashing technique used in OCaml so it is possible to write Polyvar types that act as enums within the Rust code but correctly add the hash when serializing.

An ideal API would be one using attributes at the container level e.g.

```rust
#[bin_prot(Polyvar)]
enum Example {
    #[bin_prot(label = "variant_a")]
    VariantA { a: bool },
    #[bin_prot(label = "variant_b")]
    VariantB { b: uint32 },
}
```
Each variant would be converted into a Polyvar structure
```rust
struct Polyvar<T> {
    hash: [u8; 4],
    t: T
}
```

where Hash needs to contain the hash of the label of the variant and T should be a type that can hold the variant data (e.g. a: bool or b: uint32). 

According to the OCaml implementation we need to follow the following algorithm for the hash
https://github.com/ocaml/ocaml/blob/001997e81342fd0d321fd877b73608150601e7d9/runtime/hash.c#L302 which takes a list of chars and produces an `int32`. It should read the list of chars from the label attached to the variant so we can ensure they are indentical to those used in the OCaml code.

**Specification reference**

- https://www.cs.cornell.edu/courses/cs3110/2020sp/textbook/data/polymorphic_variants.html
- https://github.com/janestreet/bin_prot#polymorphic-variants

sudo-shashank:
**Summary of changes**
- [Implemented support for strong type deserialization of Polyvar](https://github.com/ChainSafe/mina-rs/pull/191/commits/b1fafb44c307cf6c8c9d5106e2341a6ed455a504)
- [Implemented ocaml native integer support for polyvar type ser-de](https://github.com/ChainSafe/mina-rs/pull/191/commits/fa6515b80d2fb582afd4fb0791f9df74a5a73d44)
- [Added negative test case for ocaml hash variant](https://github.com/ChainSafe/mina-rs/pull/191/commits/82586e3854de6d6694c1f30389484b12a9dbb16a)
- [Added non ascii polyvar tag filter](https://github.com/ChainSafe/mina-rs/pull/191/commits/7bca28c2314207b757275385d75b4ca6fd99198d)

Closes #185 


willemolding (commented):
Looking good so far. Can it really be this easy! Just needs some tests

protocol/bin-prot/tests/polyvar.rs
```diff
@@ -0,0 +1,44 @@
...
+    assert_eq!(result, TestPolyvar::VariantTwo(true))
+}
+
+#[test]
+fn test_polyvar_unknown_polyvar_tag() {
+    let tag = 1234567_u32.to_le_bytes(); // random hash
+    let mut de = Deserializer::from_reader(tag.as_slice());
+    let result: Result<TestPolyvar, Error> = Deserialize::deserialize(&mut de);
+    assert!(result.is_err())
+}
```
creativcoder, line 44: Would be good to have a test that contains a variant which itself is a Polyvar

willemolding (changes requested):
Looking great, there is really just the one issue about the native integer format. I hope I was able to make it clear, if not I'm happy to jump on a call to try and explain

protocol/bin-prot/src/de.rs
```diff
@@ -344,6 +353,51 @@ impl<'de, 'a, R: Read> de::Deserializer<'de> for &'a mut Deserializer<R, Strongl
...
+            {
+                for (index, value) in self.variants.into_iter().enumerate() {
+                    // Return the first polyvar variant where the hash matches
+                    if caml_hash_variant(value.as_bytes()) == self.hash {
+                        let de: U8Deserializer<Self::Error> = (index as u8).into_deserializer();
+                        let v = seed.deserialize(de)?;
+                        return Ok((v, Enum::new(self.de, index as u8)));
+                    }
+                }
+                Err(Error::UnknownPolyvarTag(self.hash))
```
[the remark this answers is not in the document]
sudo-shashank: nice, thnks

protocol/bin-prot/src/polyvar/mod.rs
```diff
@@ -0,0 +1,19 @@
...
+
+/// A polymorphic variant as it should be serialized, the variant hash followed by the body.
+/// The default implementations of Serialize and Deserialize work for this.
+#[derive(Serialize, Deserialize)]
+pub struct PolyVar<T> {
+    /// The variant hash computed by using `caml_hash_variant` on its label
+    hash: VariantHash,
+    /// The body associated with the variant. Can be empty.
+    body: T,
+}
```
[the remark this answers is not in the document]
sudo-shashank: okay

protocol/bin-prot/src/de.rs
```diff
@@ -306,15 +307,23 @@ impl<'de, 'a, R: Read> de::Deserializer<'de> for &'a mut Deserializer<R, Strongl
...
         visitor: V,
     ) -> Result<V::Value>
     where
         V: Visitor<'de>,
     {
-        let index = self.rdr.bin_read_variant_index()?;
-        visitor.visit_enum(Enum::new(self, index))
+        match name {
+            "Polyvar" => {
+                let hash = self.rdr.bin_read_polyvar_tag()?;
```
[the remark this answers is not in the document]
sudo-shashank: fixed

protocol/bin-prot/src/de.rs
```diff
@@ -306,15 +307,23 @@ impl<'de, 'a, R: Read> de::Deserializer<'de> for &'a mut Deserializer<R, Strongl
...
+        variants: &'static [&'static str],
         visitor: V,
     ) -> Result<V::Value>
     where
         V: Visitor<'de>,
     {
-        let index = self.rdr.bin_read_variant_index()?;
-        visitor.visit_enum(Enum::new(self, index))
+        match name {
+            "Polyvar" => {
```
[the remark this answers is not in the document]
sudo-shashank: added comment

protocol/bin-prot/src/de.rs
```diff
@@ -344,6 +355,55 @@ impl<'de, 'a, R: Read> de::Deserializer<'de> for &'a mut Deserializer<R, Strongl
...
+            where
+                V: de::DeserializeSeed<'de>,
+            {
+                // Return the first polyvar variant where the hash matches
+                let index = self
+                    .variants
+                    .into_iter()
+                    .enumerate()
+                    .find_map(|(index, value)| {
+                        if caml_hash_variant(value.as_bytes()) == self.hash {
```
willemolding, line 390: Better make sure that value is ascii `[is_ascii](https://doc.rust-lang.org/std/string/struct.String.html#method.is_ascii)(&self) -> [bool](https://doc.rust-lang.org/std/primitive.bool.html)` otherwise computing the caml hash is undefined behaviour. We can throw an error in the case of a non-ascii tag

willemolding (approved):
One last little thing on checking the tags but I'll approve anyway. Amazing stuff!

protocol/bin-prot/src/de.rs
sudo-shashank: fixed by adding a filter on the iterator `.filter(|v| v.is_ascii())`

willemolding: My issue with this is it will throw the wrong kind of error `UnknownPolyvarTag`. I think it would be better to check and throw an explicit `PolyvarTagNotAscii` error to help with figuring out what is going on

protocol/bin-prot/src/de.rs
```diff
@@ -344,6 +355,54 @@ impl<'de, 'a, R: Read> de::Deserializer<'de> for &'a mut Deserializer<R, Strongl
...
+                        index = Some(i);
+                    }
+                }
+                match index {
+                    Some(i) => {
+                        let de: U8Deserializer<Self::Error> = (i as u8).into_deserializer();
+                        let v = seed.deserialize(de)?;
+                        Ok((v, Enum::new(self.de, i as u8)))
+                    }
+                    None => Err(Error::UnknownPolyvarTag(self.hash)),
```
sudo-shashank, line 397: @willemolding since we cannot use `caml_hash_variant?` inside `find_map`, I had to change the previous code to this

we could have used something like 
```
let index = self
       .variants
       .into_iter()
       .enumerate()
       .find_map(|(index, value)| {
           if caml_hash_variant(value).unwrap() == self.hash {
                 Some(index)
            } else {
                 None
            }
       })
       .ok_or(Error::UnknownPolyvarTag(self.hash))?;
```
but using unwrap/expect inside a lib is not a good idea

willemolding: You could expand out the `?` operator and it works quite nicely

```rust
                let index = self
                    .variants
                    .into_iter()
                    .enumerate()
                    .find_map(|(index, value)| {
                        match caml_hash_variant(value) {
                            Ok(hash) => {
                                if hash == self.hash {
                                    Some(Ok(index))
                                } else {
                                    None
                                }
                            }
                            Err(e) => Some(Err(e))
                        }
                    })
                    .ok_or(Error::UnknownPolyvarTag(self.hash))??;
```

sudo-shashank: thanks, this is perfect

creativcoder: Hey but won't this return early because we get an error being returned from find_map which is Some(err)?

sudo-shashank: yeah this will throw an error as soon as a non ascii polyvar tag is encountered

hanabi1224 (approved):
Wonderful work! Curious if we can turn some of the block deserialization tests back on with this change?

protocol/bin-prot/src/de.rs
sudo-shashank: @creativcoder @willemolding code suggestion
any suggestion for making it cleaner
```
                let mut index = None;
                let mut error = None;
                for (i, v) in self.variants.into_iter().enumerate() {
                    // Return the first polyvar variant where the hash matches
                    match caml_hash_variant(v) {
                        Ok(val) => {
                            if val == self.hash {
                                index = Some(i)
                            }
                        }
                        Err(err) => {
                            error = Some(err);
                        }
                    }
                }
                match index {
                    Some(i) => {
                        let de: U8Deserializer<Self::Error> = (i as u8).into_deserializer();
                        let v = seed.deserialize(de)?;
                        Ok((v, Enum::new(self.de, i as u8)))
                    }
                    None => match error {
                        Some(err) => Err(err),
                        None => Err(Error::UnknownPolyvarTag(self.hash)),
                    },
                }
```

sudo-shashank: or we could take out ascii check from ocaml_hash_variant method and do 
```
               let mut error = None;
                // Return the first polyvar variant where the hash matches
                let index = self
                    .variants
                    .into_iter()
                    .enumerate()
                    .find_map(|(index, value)| {
                        if !value.is_ascii() {
                            error = Some(Error::PolyvarTagNotAscii);
                            None
                        } else if caml_hash_variant(value) == self.hash {
                            Some(index)
                        } else {
                            None
                        }
                    })
                    .ok_or(Error::UnknownPolyvarTag(self.hash))?;
                let de: U8Deserializer<Self::Error> = (index as u8).into_deserializer();
                let v = seed.deserialize(de)?;
                Ok((v, Enum::new(self.de, index as u8)))
```

creativcoder: The second looks clean, but where do we use the `error` after we set it to Some(err)

creativcoder (approved):
:heavy_check_mark:

willemolding:
> Wonderful work! Curious if we can turn some of the block deserialization tests back on with this change?

Once #199 is complete then yes!

jspada (approved):
LGTM!

PR #197, opened 2022-03-31
Replace internal PublicKey type with one from proof-systems

willemolding:
# Replace internal PublicKey type with one from proof-systems

This PR is a big tidy up. It removes one placeholder types and tidies up a bunch of derives on our internal types. Apologies for the many small line changes (most are just removing derives).

Also note this references a fork of `proof-systems` that adds some helpful derives to their `CompressedPubKey` type. I've submitted a PR back to the upstream to add it. Without these derives most of our tests can't work. 

**Summary of changes**
Changes introduced in this pull request:
- Remove our placeholder `PublicKey` type
- Add conversions from the serialization types `PublicKeyV1` and `PublicKey2V1` to the proof-systems type `CompressedPubKey` 
- Remove unnecessary derives of `Serialize`, `Deserialize` and `Default` from internal types. These are actually potentially harmful as:
    - some types don't have sensible defaults (e.g what does a default public key mean?) 
    - serialization should go through the serialization types or the results will be wrong
    - Also, CompressedPubKey does not implement these derives so any containing types can't automatically derive them
- Update some tests to make less use of default
- Update the `UserCommandBuilder` to also not make use of defaults
- Remove a flaky test in `consensus/tests/common.rs` that depended on the placeholder hashing algorithm (which changed when the data structures changed)

**Reference issue to close (if applicable)**

Progress on #193 (PubKey only, still need to do Signature)

**Other information and links**

PR #198, opened 2022-03-31
M4 - Implement Hashable for num types in Account

hanabi1224:
**Summary of changes**
- Implement `mina_hasher::Hashable` for the number fields of `Account`
- Parity tests with OCaml
- Remove some stale usage of proof-system libs that do not compile with the latest version
- Adds a shim crate to manage crates from proof-systems altogether to make sure crate versions do not collide with each other

This supersedes https://github.com/ChainSafe/mina-rs/pull/162

Upstream PR(s):
https://github.com/o1-labs/proof-systems/pull/486

Closes 

### DIFF
--- a/base/src/account/mod.rs
+++ b/base/src/account/mod.rs
@@ -63,10 +63,9 @@ impl mina_hasher::Hashable for Account {
         let mut roi = ROInput::new();
         roi
             // .append_hashable(self.public_key)
-            // FIXME: avoid doing clone here once https://github.com/o1-labs/proof-systems/pull/486 is merged
-            .append_hashable(self.token_id.clone())
+            .append_hashable(&self.token_id)
             // .append_hashable(self.token_permissions)
-            .append_hashable(self.balance)
+            .append_hashable(&self.balance)
             // .append_hashable(self.nonce)
             // .append_hashable(self.receipt_chain_hash)
             // .append_hashable(self.delegate)
--- a/base/src/account/mod.rs
+++ b/base/src/account/mod.rs
@@ -8,6 +8,8 @@ pub mod permissions;
 pub mod timing;
 pub mod token_permissions;
 
+use proof_systems::*;
+
 use mina_hasher::ROInput;
 pub use permissions::{AuthRequired, Permissions};
 pub use timing::Timing;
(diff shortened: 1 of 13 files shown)


proof-systems-shim/src/lib.rs
```diff
@@ -0,0 +1,18 @@
...
+//! depending on crates from proof systems
+//!
+//! ```
+//! use proof_systems::*;
+//! use mina_hasher::Hashable;
+//! ```
+//!
+
+pub use commitment_dlog;
+pub use mina_hasher;
```
willemolding, line 18: This is a good idea, It might add another dep if people want to use our crates standalone but I think it is worth it. Also not sure if shim is the right name since it isn't modifying the API at all. I don't have any better ideas though

base/src/account/mod.rs
```diff
@@ -52,3 +55,31 @@ pub struct Account {
...
+            .append_hashable(&self.token_id)
+            // .append_hashable(self.token_permissions)
+            .append_hashable(&self.balance)
+            // .append_hashable(self.nonce)
+            // .append_hashable(self.receipt_chain_hash)
+            // .append_hashable(self.delegate)
+            // .append_hashable(self.voting_for)
+            // .append_hashable(self.timing)
+            // .append_hashable(self.permissions)
+            // .append_hashable(self.snapp)
```
willemolding, line 77: Once #197 is merged I think there are no further blockers for this right?

willemolding (approved):

base/src/account/mod.rs
hanabi1224: Yeah I believe so

proof-systems-shim/src/lib.rs
hanabi1224: Regarding naming, shim is only part of the directory name but the crate is named as `proof-systems`

base/src/account/mod.rs
```diff
@@ -52,3 +55,31 @@ pub struct Account {
     /// TODO: This should contain a Snapp account data once we have something to test against
     pub snapp: Option<()>,
 }
+
+impl mina_hasher::Hashable for Account {
+    type D = ();
+
+    // Uncomment these fields once they have implemented Hashable trait
```
creativcoder, line 62: good to prefix with a TODO, makes them easy to find

creativcoder (approved):
:heavy_check_mark:

PR #205, opened 2022-04-08
M5 - Make mvp demo mpa(multi-page application)

hanabi1224:
**Summary of changes**
- Converts the single-page demo to multi-page so that more functionalities can be integrated, e.g. readme, send transaction, etc.
- Integrate with tailwindcss
- Make chrome extension without sandboxing (requires chrome canary for now)
- Render project readme.md on index page
- Create page skeleton for graphql demo
- Build chrome extension mv2 for chrome stable / edge and firefox
- Instructions on using the extension chrome canary/stable, edge and firefox

Closes 


willemolding:
Any idea how much effort would be required to support Firefox? As I understand:
- Firefox extensions [don't support manifest v3](https://blog.mozilla.org/addons/2021/05/27/manifest-v3-update/) so we would need to switch to v2
- I think Firefox supports `unsafe-wasm-eval` in extensions (need to confirm this)

There is no Chrome Canary for Linux, so if we continue down the current path we will only support non-linux chrome users.

hanabi1224:
@willemolding 

That's a good suggestion, I've added extension mv2 support.

> There is no Chrome Canary for Linux

According to posts in the Chrome mail list, this is a bug fix and will roll out to chrome stable soon.  MV3 is required for publishing to the Chrome extension store, which is the plan made by our partner

sudo-shashank (approved):

Issue #130, opened 2022-01-20
[OB5] Ensure merkle tree Hasher and Merger implementation is compatible with mina [M4]


noot:
**Task summary**

- to ensure that our merkle tree implementation is compatible with mina's, we should test them by adding the same values to both trees, and ensuring they come to the same root hash
- we can do this by getting test vectors for their tree (either they give them to us, or we do it ourselves) and adding tests for them

* [X] Merger implementation
* [x] Hasher implementation
*

PR #208, opened 2022-04-19
M4 - Build merkle tree from genesis ledger and verify root hash

hanabi1224:
**Summary of changes**
- Add fixed-height support to `MinaMerkleLedger` (Mina uses fixed height 20 no matter how many accounts a tree stores, and the merkle root is the hash at height 20)
- Properly calculate root hash when fixed-height is set
- Verify the root hash of the Merkle tree built from the genesis ledger

This closes the merger part of https://github.com/ChainSafe/mina-rs/issues/130

Closes 


merkle/src/tree_impl.rs
```diff
@@ -119,7 +127,25 @@ where
...
+        let hash = self.calculate_hash_if_needed(0);
+        if let Some(fixed_height) = self.fixed_height {
+            match fixed_height.cmp(&self.height) {
+                Ordering::Less => panic!(
+                    "fixed_height {} should not be smaller than current height {}",
+                    fixed_height, self.height,
+                ),
+                Ordering::Equal => hash,
+                Ordering::Greater => {
+                    let mut hash = hash;
```
creativcoder, line 139: can make hash mut at `130` instead

creativcoder (approved):
:+1:

merkle/src/tree_impl.rs
hanabi1224: Fixed

willemolding (approved):

merkle/src/tree_impl.rs
```diff
@@ -45,6 +46,13 @@ where
         }
     }
 
+    /// Sets the fixed height of the merkle tree
+    /// Note that leaf nodes are at height 0 here but -1 in mina OCaml implementation
+    pub fn with_fixed_height(mut self, height: u32) -> Self {
```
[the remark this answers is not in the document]
hanabi1224: That's a great suggestion, thx!

Issue #196, opened 2022-03-31
[Tech-Debt] Refactor consensus tests in tests/common.rs and tests/density.rs


willemolding:
**Task summary**

The tests in these files are extremely verbose, contain a lot of duplication and are difficult to read (lots of mutable variables).
Ideally they should be refactored to be idiomatic and DRY, possibly with some macros to reduce duplication. 

**Specification reference**

- tests/common.rs  tests/density.rs in the consensus crate

**Other information and links**

willemolding:
Also see this https://github.com/ChainSafe/mina-rs/blob/b0d9f082515fedfc748a209044817cfb8ef2afc7/consensus/tests/common.rs#L130

PR #211, opened 2022-04-20
Implemented Hashable trait for ProtocolState

### LINKED ISSUE #166: [M2] Implement Poseidon Hash for State Hash
**Issue summary**
Implement Poseidon Hash for State Hash, Currently it uses blake2b Hash, also add test case to validate tiebreaker logic in select longer chain method

https://github.com/ChainSafe/mina-rs/blob/4d2a01611f1bed2638289a1c5751d56239bf9769/consensus/src/common.rs#L100-L104

API and framework for Mina hashing. It is a safe wrapper around Mina’s instances of the [Poseidon arithmetic sponge](https://github.com/o1-labs/cryptography-rfcs/blob/master/mina/001-poseidon-sponge.md) that converts it from a sponge into a hash interface.

sudo-shashank:
**Summary of changes**
- [Implemented hashable trait for ProtocolState and Used Kimchi Hasher](https://github.com/ChainSafe/mina-rs/pull/211/commits/0d6b5aa7e963c994e0601dad4540a06ac6fb226a)
- [Added select longer chain tie breaker logic test which use state hash comparison](https://github.com/ChainSafe/mina-rs/pull/211/commits/0d6b5aa7e963c994e0601dad4540a06ac6fb226a)

Closes #166 

### DIFF
--- a/base/src/consensus_state.rs
+++ b/base/src/consensus_state.rs
@@ -9,7 +9,8 @@ use crate::{
     numbers::{Amount, GlobalSlotNumber, Length},
 };
 use derive_more::From;
-use mina_crypto::{hash::*, prelude::*};
+use mina_crypto::prelude::*;
+use proof_systems::mina_hasher::{Hashable, ROInput};
 use proof_systems::mina_signer::CompressedPubKey;
 use serde::Serialize;
 use smart_default::SmartDefault;
@@ -18,6 +19,20 @@ use smart_default::SmartDefault;
 #[derive(Clone, Default, PartialEq, Debug, From, Serialize)]
 pub struct VrfOutputTruncated(pub Vec<u8>);
 
+impl Hashable for VrfOutputTruncated {
+    type D = ();
+
+    fn to_roinput(&self) -> ROInput {
+        let mut roi = ROInput::new();
+        roi.append_bytes(&self.0);
+        roi
+    }
+
+    fn domain_string(_: Self::D) -> Option<String> {
+        None
+    }
+}
+
 impl Base64Encodable for VrfOutputTruncated {}
 
 impl From<&str> for VrfOutputTruncated {
@@ -25,8 +40,6 @@ impl From<&str> for VrfOutputTruncated {
         VrfOutputTruncated(s.as_bytes().to_vec())
     }
 }
-
-impl Hashable<VrfOutputHash> for VrfOutputTruncated {}
 
 impl AsRef<[u8]> for VrfOutputTruncated {
     fn as_ref(&self) -> &[u8] {
@@ -78,6 +91,36 @@ pub struct ConsensusState {
     pub coinbase_receiver: CompressedPubKey,
     /// true if block_stake_winner has no locked tokens, false otherwise
     pub supercharge_coinbase: bool,
+}
+
+impl Hashable for ConsensusState {
+    type D = ();
+
+    fn to_roinput(&self) -> ROInput {
+        let mut roi = ROInput::new();
+        roi.append_hashable(&self.blockchain_length);
+        roi.append_hashable(&self.epoch_count);
+        roi.append_hashable(&self.min_window_density);
+        for v in &self.sub_window_densities {
+            roi.append_hashable(v);
+        }
+        roi.append_hashable(&self.last_vrf_output);
+        roi.append_hashable(&self.total_currency);
+        roi.append_hashable(&self.curr_global_slot);
+        roi.append_hashable(&self.global_slot_since_genesis);
+        roi.append_hashable(&self.staking_epoch_data);
+        roi.append_hashable(&self.next_epoch_data);
+        roi.append_bool(self.has_ancestor_in_same_checkpoint_window);
+        roi.append_bytes(self.block_stake_winner.into_address().as_bytes());
+        roi.append_bytes(self.block_creator.into_address().as_bytes());
+        roi.append_bytes(self.coinbase_receiver.into_address().as_bytes());
+        roi.append_bool(self.supercharge_coinbase);
+        roi
+    }
+
+    fn domain_string(_: Self::D) -> Option<String> {
+        None
+    }
 }
 
 impl ConsensusState {
--- a/consensus/src/common.rs
+++ b/consensus/src/common.rs
@@ -6,12 +6,16 @@
 //!
 
 use crate::error::ConsensusError;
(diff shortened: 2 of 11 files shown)


consensus/src/common.rs
```diff
@@ -120,21 +121,16 @@ impl Chain<ProtocolState> for ProtocolStateChain {
...
+            hasher.hash(&s.body.consensus_state.last_vrf_output)
         })
     }
 
-    // FIXME: this currently returns a blake2b hash. It should
-    // return a poseidon hash according to: https://github.com/MinaProtocol/mina/blob/32f529d8d8d712a44ee75be66061ce08cbdc8924/docs/specs/consensus/README.md#517-hashstate
-    fn state_hash(&self) -> Option<StateHash> {
-        self.top().map(|s| s.hash())
+    fn state_hash(&self) -> Option<Fp> {
+        let mut hasher = create_kimchi::<ProtocolState>(());
```
hanabi1224, line 132: It's expensive to create a new hasher every time, better to create a pool and reuse the hasher.

hanabi1224 (approved):
Nice work! Approved with some suggestions

proof-systems-shim/src/lib.rs
```diff
@@ -14,6 +14,7 @@
 //! ```
 //!
 
+pub use mina_curves;
```
hanabi1224, line 17: You don't need to change this crate, Fp is re-exported from mina_hasher

proof-systems-shim/Cargo.toml
```diff
@@ -7,6 +7,8 @@ edition = "2021"
 license = "Apache-2.0"
 
 [dependencies]
-mina-hasher = {git = "https://github.com/o1-labs/proof-systems", rev = "9816b804fe2e8336fbe20d2e3d977d766582c4f0"}
```
hanabi1224, line 10: Please also revert the changes here, `Fp` is available under `mina-hasher`

proof-systems-shim/src/lib.rs
sudo-shashank: fixed

base/src/consensus_state.rs
```diff
@@ -80,6 +93,36 @@ pub struct ConsensusState {
     pub supercharge_coinbase: bool,
 }
 
+impl Hashable for ConsensusState {
```
creativcoder, line 96: This impl block is mostly same except for the function body of `to_roinput`. Can we write a readable macro to remove the repetitive parts? I believe you could just use a token tree `tt` for the function body.

base/src/consensus_state.rs
```diff
@@ -18,6 +20,11 @@ use smart_default::SmartDefault;
 #[derive(Clone, Default, PartialEq, Debug, From, Serialize)]
 pub struct VrfOutputTruncated(pub Vec<u8>);
 
+impl_ref_hashable!(
+    VrfOutputTruncated,
+    append_bytes => 0
```
willemolding, line 25: This macro is cute but I think it makes things less readable rather than more. It took me a while to figure out what is going on in this one for example.

Also it doesn't really save much space. I think this would be better without it

proof-systems-shim/src/lib.rs
```diff
@@ -17,3 +17,43 @@
 pub use mina_hasher;
 pub use mina_signer;
 pub use o1_utils;
+
+/// macro that implements hashable trait for a reference value
+#[macro_export]
```
willemolding, line 22: If we do have to include this macro I think it should live somewhere else. I think we should keep the shim crate as a pure shim (no additional code)

consensus/src/common.rs
```diff
@@ -4,12 +4,16 @@
 //! Implements common APIs for the blockchain in the context of consensus.
```
willemolding, line 4: This adds a bunch more of the difficult-to-read, repeated code tests we are trying to remove in #196. I don't think this is the way we should be writing these. Maybe chat with @creativcoder to see how they are planning to rewrite them and we can do it that way in the first place

willemolding (changes requested):
Main point is to improve the strategy for writing tests in `consensus/src/common.rs`

base/src/consensus_state.rs
sudo-shashank: removed macro use for now

consensus/src/common.rs
creativcoder: Yes, we need not manually populate test payloads this way. For the same reason, I wanted to have our base internal types i.e., `ProtocolState` annotated with `Serialize`, `Deserialize`? @willemolding is it possible for us to add the Serialize annotation to those; would they break anything?

consensus/src/common.rs
```diff
@@ -120,21 +124,23 @@ impl Chain<ProtocolState> for ProtocolStateChain {
...
-        self.top().map(|s| {
-            s.body
-                .consensus_state
-                .last_vrf_output
-                .hash()
-                .as_ref()
-                .encode_hex::<String>()
-        })
+    fn last_vrf_hash(&self) -> Option<Fp> {
+        static HASHER_POOL: OnceCell<SpinLockObjectPool<PoseidonHasherKimchi<VrfOutputTruncated>>> =
```
creativcoder, line 128: Let's move this pool initialization block at the top as a function which returns `pool`. I see we initialize pool both in `state_hash()` and `last_vrf_hash()`

creativcoder, line 128: Let's move this pool initialization block at the top as a function which returns `pool`. I see we initialize pool both in `state_hash()` and `last_vrf_hash()`

sudo-shashank: hasher type is different for both, `state_hash()` uses ProtocolState and `last_vrf_hash()` uses VrfOutputTruncated

creativcoder (approved):

proof-systems-shim/src/lib.rs
creativcoder: Removed the macro as was not really helping much

consensus/src/common.rs
creativcoder: @willemolding  Update: We decided to remove this test and add it later as part of https://github.com/ChainSafe/mina-rs/issues/196 Additionally for removing the tech debt we'll need to have json serialization support from base types, which can be added after https://github.com/ChainSafe/mina-rs/pull/209 is merged. 
So merging this now.

PR #212, opened 2022-04-21
M1 - Add TypeAnnotation trait to support multiple serde formats (binprot and json)

hanabi1224:
**Summary of changes**

This is part of https://github.com/ChainSafe/mina-rs/pull/209 it contains the minimal changes that add `TypeAnnotation` trait, to make review easier and unblock other's work that rely on this trait.

- Add `TypeAnnotation` trait

Closes https://github.com/ChainSafe/mina-rs/issues/213

### DIFF
--- a/apps/profilers/src/bin/block-serde-profiler.rs
+++ b/apps/profilers/src/bin/block-serde-profiler.rs
@@ -31,7 +31,7 @@ fn main() -> anyhow::Result<()> {
 
 fn cpu_profile_serialization() -> anyhow::Result<ExternalTransition> {
     let mut de = bin_prot::Deserializer::from_reader(BLOCK_BYTES);
-    let et: <ExternalTransition as SerializationTypeAnnotation>::BinProtType =
+    let et: <ExternalTransition as BinProtSerializationType>::T =
         serde::Deserialize::deserialize(&mut de)?;
     Ok(et.into())
 }
--- a/apps/profilers/src/bin/block-serde-profiler.rs
+++ b/apps/profilers/src/bin/block-serde-profiler.rs
@@ -1,9 +1,8 @@
 // Copyright 2020 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0
 
-use bin_prot::encodable::BinProtEncodable;
 use clap::{Arg, Command};
-use mina_rs_base::types::ExternalTransition;
+use mina_rs_base::{types::*, *};
 use std::str::FromStr;
 
 const BLOCK_BYTES: &[u8] = include_bytes!("../../../../protocol/test-fixtures/src/data/block1");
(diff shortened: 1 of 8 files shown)


protocol/serialization-types/src/type_annotation.rs
```diff
@@ -0,0 +1,16 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0
+
+/// This trait annotates a given type its different serialization types,
+/// and provide utility functions to easily convert between them
+pub trait TypeAnnotation: Sized {
```
creativcoder, line 6: Can we name this trait to `DataFormat`, like how they mention of data formats in serde: https://docs.rs/serde/latest/serde/#data-formats

creativcoder (approved):

protocol/serialization-types/src/type_annotation.rs
willemolding: I suggest we rename the trait to `SerializationTypes`

willemolding: and the associated types to just `BinProt` and `Json`

apps/profilers/src/bin/block-serde-profiler.rs
```diff
@@ -31,7 +30,10 @@ fn main() -> anyhow::Result<()> {
 }
 
 fn cpu_profile_serialization() -> anyhow::Result<ExternalTransition> {
-    Ok(ExternalTransition::try_decode_binprot(BLOCK_BYTES)?)
+    let mut de = bin_prot::Deserializer::from_reader(BLOCK_BYTES);
+    let et: <ExternalTransition as SerializationTypeAnnotation>::BinProtType =
```
willemolding, line 34: This is very verbose but would be just `<ExternalTransition as SerializationType>::BinProt` with suggest renaming

protocol/serialization-types/src/type_annotation.rs
```diff
@@ -0,0 +1,16 @@
...
+    /// The corresponding serialization type for bin-prot format
+    /// Self type can be used here to indicate no special convertion is needed
+    /// TODO: Add default value when the feature lands on stable rust
+    type BinProtType: From<Self> + Into<Self>;
+
+    /// The corresponding serialization type for json format
+    /// Self type can be used here to indicate no special convertion is needed
+    /// TODO: Add default value when the feature lands on stable rust
+    type JsonType: From<Self> + Into<Self>;
+}
```
willemolding, line 16: Another suggestion would be to split the two types into different traits. This would allow us to express if some types should never be serialized to Json for example. In this case it would be 

```rust
pub trait JsonSerializationType {
    type T: From<Self> + Into<Self>;
}
```

willemolding (commented):
I think we need to settle on the naming and API for the trait. All round though this is an important addition that will hopefully avoid any confusion down the track

apps/profilers/src/bin/block-serde-profiler.rs
hanabi1224: fixed

protocol/serialization-types/src/type_annotation.rs
hanabi1224: fixed

willemolding (approved):
I like this a lot. We can make it very clear with by tagging with these traits which types can be serialized in which ways and how.

PR #218, opened 2022-04-22
M1 - Macro for testing binprot sum type

hanabi1224:
**Summary of changes**

Previously, we assume the enum variant of a sum type in tests, which is bad and unscalabe when more than one variant appear in the test blocks.   This PR adds a new macro for testing sum types by asserting exactly one variant get deserializaed into successfully.

- 

Closes 


creativcoder (approved):

willemolding (approved):
So good!!

PR #224, opened 2022-05-04
Add ExtendedInteger types

### LINKED ISSUE #206: [M1] Implement ExtendedU32 correctly
**Task summary**

Mina uses its own unsigned numeric data type `Nat.UInt32` for  a number of internal types including `AccountNonce` and `GlobalSlotNumber`.

This has slightly different serialization to a bin_prot integer from what I can tell. For the most part they are compatible but in some edge cases they are not.

Our codebase uses i32 and coverts to u32 in order to maintain compatibility. This is a hack solution and can fail in certain cases (which are not covered properly by tests)

Tasks for this issue

- Explicitly support `Nat.UInt32` as a serialization type. Currently we use `ExtendedU32` wrapper around i32
- Add conversions from this to rust types (e.g. u32)
- Replace this in our other serialization-types where required
- Add test coverage for edge cases for ExtendedU32 (e.g. minvalue, maxvalue etc)

**Specification reference**

- https://github.com/MinaProtocol/mina/blob/develop/src/lib/mina_numbers/nat.ml

willemolding:
**Summary of changes**
- Adds ExtendedU32 and ExtendedU64 types with custom deserialization
- Uses these in conversion from serialization types

Closes #206 


base/tests/serialization_type_conversions.rs
```diff
@@ -5,17 +5,17 @@ use mina_rs_base::types::ExternalTransition;
...
-        .get("block1")
-        .expect("Failed to load block1")
-        .external_transitionv1()
-        .unwrap();
-    let internal_type_block = ExternalTransition::from(serialization_type_block.clone());
-    let recovered_serialization_type_block = ExternalTransitionV1::from(internal_type_block);
+// #[test]
+// /// Test both From impls to convert between the serialization/wire-type of a block and the internal representation
+// /// The conversion must be lossless!
+// fn roundtrip_serialization_types() {
```
hanabi1224, line 11: Why is this test disabled?

hanabi1224, line 11: Why is this test disabled?

willemolding:
Since all the underlying serialization code has changed this PR is no longer relevent

PR #239, opened 2022-05-30
M4 - merkle proof implementation and sparse merkle tree trait

### LINKED ISSUE #240: M4 - Implement merkle proof
**Task summary**

https://www.webopedia.com/definitions/merkle-proof/

**Specification reference**

- 

hanabi1224:
**Summary of changes**
- Implement `MerkleProof` with unit tests
- Define `SparseMerkleTree` trait
- Remove generic `DEGREE` support from merkletree-related data structures as it add unwanted complexity

Closes #240 


merkle/src/proof.rs
```diff
@@ -0,0 +1,111 @@
...
+where
+    Hash: PartialEq + Clone + std::fmt::Debug,
+    Hasher: MerkleHasher<Item = Item, Hash = Hash>,
+    Merger: MerkleMerger<Hash = Hash>,
+{
+    type Hash = Hash;
+
+    fn root_hash(&self) -> Option<Hash> {
+        if self.index > 0 {
+            // let mut peer_index = get_peer_index(self.index);
```
creativcoder, line 79: remove this?

creativcoder, line 79: remove this?

merkle/src/tree_impl.rs
```diff
@@ -78,6 +100,64 @@ where
...
+            let height_above = if let Some(fixed_height) = self.mode.fixed_height() {
+                fixed_height
+            } else {
+                self.variable_height
+            };
+            let index_offset = calculate_node_count(height_above);
+            let capacity = self.mode.fixed_height().unwrap_or(self.variable_height) as usize;
+            let mut peer_indices = Vec::with_capacity(capacity);
+            let mut peer_hashes = Vec::with_capacity(capacity);
+            let (item, _) = &self.leafs[index];
```
creativcoder, line 119: is there a possiblity of panicking on indexing here? does `get_proof` always receives a valid index?

creativcoder (approved):
neat! left a few comments

merkle/src/proof.rs
hanabi1224: fixed

hanabi1224: fixed

merkle/src/tree_impl.rs
hanabi1224: Yes, updated comment doc.  This function is mainly for unit tests, we will be constructing proof from graphQL API directly in  the web node

merkle/src/proof.rs
```diff
@@ -0,0 +1,109 @@
...
+
+use crate::*;
+
+/// Merkle proof trait of a single leaf node, for details refer to <https://www.webopedia.com/definitions/merkle-proof/>
+pub trait MerkleProof {
+    /// Hash type
+    type Hash: PartialEq;
+
+    /// Calculates the root hash
+    fn root_hash(&self) -> Option<Self::Hash>;
```
willemolding, line 16: I'm not sure if I agree with the semantics here with the optional type. If a merkle proof doesn't have a root hash that means it isn't a merkle proof and so shouldn't be implementing this trait

merkle/src/proof.rs
```diff
@@ -0,0 +1,109 @@
...
+                let parent_index = get_parent_index(index);
+                hash_opt = Merger::merge(
+                    hashes,
+                    MerkleTreeNodeMetadata::new(parent_index, self.peer_indices.len() as u32),
+                );
+                index = parent_index;
+            }
+            hash_opt
+        } else {
+            None
```
willemolding, line 101: Similarly here, if the index is 0 this just means it is the root node right so it can just return itself?

merkle/src/tree.rs
```diff
@@ -14,9 +16,39 @@ pub trait MerkleTree<const DEGREE: usize> {
...
+    type Item;
+    /// Type of the hash values
+    type Hash: PartialEq + Clone;
+    /// Type of the merkle hasher
+    type Hasher: MerkleHasher<Item = Self::Item, Hash = Self::Hash>;
+    /// Type of the merkle merger
+    type Merger: MerkleMerger<Hash = Self::Hash>;
+
+    /// Adds a single [MerkleProof]
+    fn add(
```
willemolding, line 40: Make this generic over types that implement `MerkleProof` rather than forcing it to be a `DefaultMerkleProof`

merkle/src/tree_impl.rs
```diff
@@ -78,6 +100,65 @@ where
         }
     }
 
+    /// Gets the merkle proof of an item with the 0-based index of the item
+    /// being added, e.g. the first item is index 0.
+    /// This function panics when the index is out of range.
+    pub fn get_proof(
+        &mut self,
+        index: usize,
+    ) -> Option<DefaultMerkleProof<Item, Hash, Hasher, Merger>> {
```
willemolding, line 109: This function body is really hard to read and understand for me. I think it could benefit from some comments and possibly rearranging a bit for readibility

merkle/tests/tree.rs
```diff
@@ -159,6 +156,20 @@ mod tests {
         test_expand_mina_merkle_tree(&mut tree, 188, 17578, 8);
     }
 
+    #[test]
+    fn test_proofs() {
```
willemolding, line 160: Nice succinct way to test :)

merkle/src/proof.rs
hanabi1224: Maybe I should use result here. In current impl a proof is not validated on construction and it's possible to have nodes missing along the path and fail to calculate the root hash

merkle/src/proof.rs
hanabi1224: 0 will never be the index of a data node in current tree impl so it's not a valid value here

hanabi1224: Maybe the comment is misleading but a data node will never be root. Will try to improve that

merkle/src/tree.rs
hanabi1224: Good catch! Fixed

merkle/src/proof.rs
hanabi1224: fixed

merkle/src/proof.rs
hanabi1224: fixed

merkle/src/tree_impl.rs
hanabi1224: Added comments to describe the flow, plz let me know whether it helps

merkle/src/proof.rs
```diff
@@ -0,0 +1,150 @@
...
+                let peer_index = self.peer_indices[i];
+                let peer_hash_opt = &self.peer_hashes[i];
+                // 3 Prepare the input for for the hash merger by ordering the 2 hashes properly
+                let hashes = if index + 1 == peer_index {
+                    [hash_opt, peer_hash_opt.clone()]
+                } else if index == peer_index + 1 {
+                    [peer_hash_opt.clone(), hash_opt]
+                } else {
+                    return Err(MerkleProofError::InvalidProof);
+                };
```
willemolding, line 125: This would be great to refactor to its own function `get_ordered_siblings` or something

willemolding (approved):
Heaps more readable now for the main logic part

merkle/src/proof.rs
hanabi1224: Fixed

Issue #249, opened 2022-06-09
[M] Implementing tracing and logging


creativcoder:
**Issue summary**

**Other information and links**

PR #250, opened 2022-06-10
MVP - browser console logger and exportable logger

hanabi1224:
**Summary of changes**
- Impl `JsExportableLogger`
- Add `Export Log` button to web UI

Closes https://github.com/ChainSafe/mina-rs/issues/249


apps/wasm/mvp/src/logger.rs
```diff
@@ -0,0 +1,79 @@
+// Copyright 2020 ChainSafe Systems
+// SPDX-License-Identifier: Apache-2.0
+
+use log::{Level, Log, Metadata, Record};
+use wasm_bindgen::prelude::*;
+
+use crate::js_ffi::console_log_str;
+
+#[wasm_bindgen(inline_js = r###"
```
creativcoder, line 9: Is the string based inline js code the only way to do this? or maybe move this to a js file with the module attribute if that works?

apps/wasm/mvp/utils.ts
```diff
@@ -7,3 +7,18 @@ import init from "./pkg/wasm"
...
+    const anchor = document.createElement('a');
+    anchor.href = path;
+    anchor.download = filename;
+    anchor.style.display = 'none';
+    document.body.appendChild(anchor);
+    anchor.click();
+    document.body.removeChild(anchor);
+}
+
+export function textToPath(text, type = "text/plain") {
```
creativcoder, line 21: are we exporting the logs as newline delimited json? if yes, shouldn't we use `application/json` as mime type?

hanabi1224: I'm not sure if I understand, but the newline delimited text is not a valid json

apps/wasm/mvp/src/logger.rs
hanabi1224: Yeah we can do that but I just feel that makes it less readable since IDE does not provide code nav and it's a short code snippet. Does that make sense?

apps/wasm/mvp/src/js_ffi.rs
```diff
@@ -6,11 +6,8 @@ use wasm_bindgen::prelude::*;
 
 #[wasm_bindgen]
 extern "C" {
-    #[wasm_bindgen(js_namespace= console, js_name = log)]
-    pub fn log_string(s: String);
-
-    #[wasm_bindgen(js_namespace= console, js_name = log)]
-    pub fn log_str(s: &str);
+    #[wasm_bindgen(js_namespace = console,js_name = "log")]
+    pub fn console_log_str(s: &str);
```
sudo-shashank, line 10:
```suggestion
    pub fn console_log(s: &str);
```

sudo-shashank (approved):

apps/wasm/mvp/src/js_ffi.rs
hanabi1224: fixed

apps/wasm/mvp/src/logger.rs
creativcoder: got it, it should be fine.

apps/wasm/mvp/utils.ts
creativcoder: usually log ingestion service support: https://jsonlines.org/ But i believe the use case is trivial as of now. So it should be fine

creativcoder (approved):

PR #282, opened 2022-09-13
Refactor Chain Selection

sudo-shashank:
**Summary of changes**
- Refactored chain selection to support berkley net 

Closes 

### DIFF
--- a/base/src/blockchain_state.rs
+++ b/base/src/blockchain_state.rs
@@ -5,6 +5,7 @@
 
 use std::convert::TryInto;
 
+use crate::types::*;
 use crate::{
     blockchain_state_registers::BlockchainStateRegisters,
     numbers::{BlockTime, TokenId},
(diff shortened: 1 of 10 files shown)


base/src/blockchain_state.rs
```diff
@@ -82,6 +84,56 @@ impl ToChunkedROInput for BlockchainState {
     }
 }
 
+impl Into<BlockchainState> for BlockchainStateLegacy {
```
hanabi1224, line 87: Why Into trait? I don't think they are compatible with those breaking changes.

consensus/src/common.rs
```diff
@@ -51,10 +51,10 @@ impl ConsensusConstants {
     }
 }
 
-/// A chain of [ProtocolStateLegacy]
+/// A chain of [ProtocolState]
 #[derive(Debug, Default, Eq, PartialEq, Clone)]
 // TODO: replace vec element with ExternalTransition
-pub struct ProtocolStateChain(pub Vec<ProtocolStateLegacy>);
```
hanabi1224, line 57: Why not use generic?

base/src/blockchain_state.rs
sudo-shashank: in some chain selection test cases we directly read the block json and check, this trait is introduced to keep those compatible with new ProtocolState

PR #283, opened 2022-09-15
Refactor chain selection

hanabi1224:
**Summary of changes**
- Refactor chain selection APIs to support generic protocol state
- Refactor `NaiveTransitionFrontier` to support generic protocol state
- Fix a bug in `select_secure_chain`

Closes 


consensus/src/common.rs
```diff
@@ -103,88 +77,93 @@ impl Chain<ProtocolStateLegacy> for ProtocolStateChain {
...
     fn config(&self) -> ConsensusConstants;
 }
 
-impl Consensus for ProtocolStateChain {
-    type Chain = ProtocolStateChain;
-    fn select_secure_chain<'a>(
-        &'a self,
-        candidates: &'a [Self::Chain],
-    ) -> Result<&'a ProtocolStateChain, ConsensusError> {
-        let tip = candidates.iter().fold(Ok(self), |tip, candidate| {
```
hanabi1224, line 184: Looks like there was a bug, with `fold` fn, all `self` usage should be replaced by `tip`

consensus/src/common.rs
```diff
@@ -103,88 +77,93 @@ impl Chain<ProtocolStateLegacy> for ProtocolStateChain {
...
-    /// Chain type
-    type Chain;
+/// A trait that defines operations for chain selection
+pub trait ChainSelection {
     /// Top level API to select between chains during a fork.
-    fn select_secure_chain<'a>(
-        &'a self,
-        candidates: &'a [Self::Chain],
-    ) -> Result<&'a ProtocolStateChain, ConsensusError>;
+    fn select_secure_chain(&mut self, candidates: Vec<Self>) -> Result<(), ConsensusError>
```
hanabi1224, line 139: Using borrows is sufficient for writing tests but not for real-world cases. IMO it's better to take over the ownership to avoid unnecessary clones

hanabi1224: And it makes it much easier for users if this API takes mutable self and makes in-place updates, as is shown in `naive_transition_frontier.rs`

consensus/src/common.rs
```diff
@@ -51,45 +51,19 @@ impl ConsensusConstants {
...
 #[derive(Debug, Default, Eq, PartialEq, Clone)]
 // TODO: replace vec element with ExternalTransition
-pub struct ProtocolStateChain(pub Vec<ProtocolStateLegacy>);
+pub struct ProtocolStateChain<T>(pub Vec<T>)
+where
+    T: ProtocolStateHeader;
 
-/// Trait that represents a chain of block data structure
-pub trait Chain<T>
+impl<T> ProtocolStateChain<T>
```
willemolding, line 61: Any particular reason for refactoring these methods out of a trait?

consensus/src/common.rs
```diff
@@ -103,88 +77,93 @@ impl Chain<ProtocolStateLegacy> for ProtocolStateChain {
...
-        &'a self,
-        candidates: &'a [Self::Chain],
-    ) -> Result<&'a ProtocolStateChain, ConsensusError> {
-        let tip = candidates.iter().fold(Ok(self), |tip, candidate| {
-            if self.is_short_range(candidate)? {
+impl<T> ChainSelection for ProtocolStateChain<T>
+where
+    T: ProtocolStateHeader,
+{
+    fn select_secure_chain(&mut self, candidates: Vec<Self>) -> Result<(), ConsensusError> {
```
willemolding, line 162: Also what is the reasoning behind switching to mutating self rather than a more functional approach?

The functional way seems like a much better API to me. Is the switch to reduce copies?

willemolding (approved):
Without context its hard to understand the reasoning behind this refactor but the changes themselves look sound.

consensus/src/common.rs
hanabi1224: I don't see why we would impl this trait for other structs so IMO it's a bit over-designed.  With a trait that support generic, it requires lifetime markers that harms readability a bit

consensus/src/common.rs
hanabi1224: The transition frontier holds the best chain for the entire lifecycle of the application, thus it's the most intuitive way to make best chain an owned field in the transition frontier.  When we use borrows in the chain selection API it's hard to update the owned best chain from caller side, or when we takes an owned self and returns an owned self, it makes it hard for the caller to work with the borrow checker.   Basically, the caller has a mutable transition frontier struct, and uses a loop to run chain selection when new blocks arrive,  it's hard to do something like
```rust
let mut frontier = ...;
frontier.best_chain = frontier.best_chain.select_best_chain(...)
```

hanabi1224: The transition frontier holds the best chain for the entire lifecycle of the application, thus it's the most intuitive way to make best chain an owned field in the transition frontier.  When we use borrows in the chain selection API it's hard to update the owned best chain from caller side, or when we takes an owned self and returns an owned self, it makes it hard for the caller to work with the borrow checker.   Basically, the caller has a mutable transition frontier struct, and uses a loop to run chain selection when new blocks arrive,  it's hard to do something like
```rust
let mut frontier = ...;
frontier.best_chain = frontier.best_chain.select_best_chain(...)
```

